PR #114, opened 2026-04-23
[Feature] Sphinx Documentation - Build System

CoolCatMona:
Creates `build_system` which contains a `Dockerfile` that installs the project dependencies as well as [sphinx](https://www.sphinx-doc.org/)

Creates a `Makefile` which has macros for making the `image` and `html`

This could reasonably be put into whatever ci-cd pipeline to ensure that the documentation builds successfully


build_system/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+RUN uv pip install --system --no-cache \
+    sphinx==8.1.3 \
+    myst-parser==4.0.1 \
+    furo==2025.12.19 \
+    sphinx-autodoc-typehints==3.0.1
+
+COPY pyproject.toml /tmp/pyproject.toml
+RUN uv pip install --system --no-cache -r /tmp/pyproject.toml
+
+WORKDIR /project
```
CoolCatMona, line 14: Really simple docker image that just installs sphinx deps and project deps

Probably a better way of managing dependencies (i.e. with a requirements.txt somewhere), but this should be fine for now

docs/conf.py
```diff
@@ -0,0 +1,47 @@
+import sys
+import os
+
+sys.path.insert(0, os.path.abspath(".."))
```
CoolCatMona, line 4: Documentation from the project root so we can find things

.gitignore
```diff
@@ -39,6 +39,8 @@ Cargo.lock
 /frontend/pkg/
 /frontend/.parcel-cache/
 /frontend/node_modules/
+/docs/_build/
+*.egg-info/
```
CoolCatMona, line 45: Don't want to commit the docs to source, nor the egg

CoolCatMona:
@reid23 @Aidan-Phan :pleading_face:

docs/conf.py
```diff
@@ -0,0 +1,47 @@
...
+
+myst_enable_extensions = [
+    "deflist",
+    "colon_fence",
+]
+
+# Auto-generate heading anchors so that same-page fragment links resolve correctly.
+myst_heading_anchors = 4
+
+exclude_patterns = ["_build", "README.md"]
```
reid23:
```suggestion
exclude_patterns = ["_build", "README.md", "uploads/videos/**"]
```

on the dev server the videos folder can be quite large from testing, so no need to copy it

reid23 (changes requested):
literally just one small change to exclude the videos, otherwise this looks good

I've updated the dev server update script to build these docs, so every time things are pushed to `dev`, the docs will update. they're live at `arctos-docs.cftocp.gay`

PR #115, opened 2026-04-23
[Feature] Add missing docstrings

CoolCatMona:
Generate docstrings with type hints for code base

IDEs like this

Not gonna lie though I had my friend claude do these, so im hoping you agree/disagree that these are about correct.


CoolCatMona:
@reid23 @Aidan-Phan

reid23:
yeah im in no position to judge here

PR #143, opened 2026-04-24
Enable Pragma, Setup Alembic Baseline

### LINKED ISSUE #90: deletion of player and team accounts
if a player or team has any active records, we don't want to delete the account (we cannot since that would cause tons of issues with missing data). this should show a reasonable error.

### LINKED ISSUE #137: Enable PRAGMA, Establish Alembic Baseline
| Step | Action |
|---|---|
| 0.1 | Add `PRAGMA foreign_keys = ON` to the SQLAlchemy `connect` listener. This is a one-line change with no migration needed. |
| 0.2 | Establish an Alembic baseline (`alembic stamp head`) so all subsequent phases produce proper versioned migrations. |
| 0.3 | Add a pre-migration backup script: `sqlite3 tournament.db ".backup backups/pre_phase1_$(date +%s).db"`. Run it before every subsequent phase. |
| 0.4 | Add the duplicate-check queries from to CI or a one-off script, so you know the data is clean before adding unique indexes. |

- [x] The `set_sqlite_pragmas` event listener function in `app/__init__.py` has a one-line comment explaining *why* FK enforcement is enabled (SQLite disables it by default — this is non-obvious).
- [x] The backup script in `scripts/` has a module-level docstring with usage instructions and a note about when it must be run.

CoolCatMona:
Enforce Foreign Key Constraints at Runtime
- app/__init__.py: replaced the old _sqlite_pragma listener with a properly named set_sqlite_pragmas. It now applies journal_mode=WAL, busy_timeout=30000, and foreign_keys=ON. The function carries a long docstring spelling out why each pragma is set

Enable Alembic Baseline for future migrations
- Added alembic to dependencies
- New alembic.ini at the repo root with ruff post-write hook so generated migrations stay formatted.
- migrations/env.py boots without invoking create_app() (avoids boot-time schedule recompute and other side effects), resolves sqlite:///instance/tournament.db the same way Flask does, and forces PRAGMA foreign_keys = ON on the migration connection so DDL is checked the same way runtime is.
- migrations/script.py.mako enforces a real docstring placeholder for every generated migration.
- migrations/versions/0001_baseline.py is an explicit, well-documented no-op so existing databases can be brought under Alembic with alembic stamp head.
- migrations/README.md documents the day-to-day workflow.
- Makefile gained db-baseline, db-migrate, db-revision MSG=…, db-current, db-history.

Created backup script
- scripts/backup_db.sh uses the SQLite online .backup (per the doc), writes to backups/<basename>-<phase>-<unix_ts>.db, runs an integrity_check on the snapshot, refuses to clobber, and exits non-zero with a clear message if sqlite3 is missing.
- The module-level header explicitly states purpose, when to run it, what it does, configuration knobs, exit codes, and the recovery procedure (including removing the WAL/SHM sidecars).
- Added sqlite3 / sqlite to setup/apt-packages.txt and the Brewfile so fresh dev machines have it; added backups/ to .gitignore. Wired make db-backup.
Modified and moved tests to get things to work at this stage (since foreign keys werent enforced)

Duplicate Check Script
- scripts/check_duplicates.py runs every query declaratively from a typed DuplicateCheck table. 
= Skips tables that don't exist (so it works against fresh DBs), excludes NULL groups (which SQLite's UNIQUE allows and would be false positives), prints a copy-pasteable report, returns 0/1 for CI use, and accepts --db <path> or SQLALCHEMY_DATABASE_URI. Wired make db-check-duplicates.

Fixed / Move some tests
- Missing parent rows. Tests inserted Match.team1='team1', TeamRegistration.team='t1', etc. without ever creating the corresponding Team rows. Fixed by adding a seeded_teams opt-in fixture in tests/conftest.py that seeds the dummy team IDs legacy tests reference; failing tests now declare it.
- Insert-order race. Tests like test_permission_service_can_view_unpublished_tournament_for_to added a Tournament and a TO(event=t.url) in the same commit. SQLAlchemy's unit of work only orders inserts via relationship() declarations — bare ForeignKey columns establish no dependency — so SA was emitting the TO insert first, which now (correctly) fails. Fixed with explicit db.session.flush() calls between the parent and child adds.

### DIFF
--- a/setup/apt-packages.txt
+++ b/setup/apt-packages.txt
@@ -1,3 +1,2 @@
 git
-git-lfs
 sqlite3
--- a/setup/apt-packages.txt
+++ b/setup/apt-packages.txt
@@ -1,2 +1,3 @@
 git
 git-lfs
+sqlite3
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ instance/*
 static/uploads/*
 preview/*
 dioxus/
+backups/
 
 
 # certs
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,9 @@ CERT_SUBJECT ?= /CN=localhost
         setup setup-os setup-macos setup-ubuntu setup-python install setup-frontend \
         lint format \
         test unit integration \
-        run certs
+        run certs \
+        db-baseline db-migrate db-revision db-current db-history \
+        db-backup db-check-duplicates
 
 help: ## Show available targets
 	@awk 'BEGIN {FS = ":.*##"; printf "Usage: make <target>\n\nTargets:\n"} \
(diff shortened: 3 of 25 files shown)


setup/apt-packages.txt
```diff
@@ -1,2 +1,3 @@
 git
-git-lfs
\ No newline at end of file
+git-lfs
```
reid23: why do we need `git-lfs`? i dont see anything that big that we're tracking rn and tbh i think if we have big files that require it we should probably use something other than git to track them so they dont make this repo super slow?

reid23 (approved):
I've never used alembic before but so far from what I've read it looks great. much needed.

i left a comment about git-lfs but even if its not needed idc so this is marked approved.

setup/apt-packages.txt
CoolCatMona: I'll remove it :)

Usually its used for larger files that dont change often.

PR #144, opened 2026-04-24
Additive Changes to Database Schema

### LINKED ISSUE #138: Create new tables, indexes, address type fixes
**New normalised tables**

Create all new tables

**Fix monetary column types**

Use `batch_alter_table` for each of the four columns. SQLite does not support
`ALTER COLUMN` natively; Alembic's batch mode copies the table.

```python
# In the Alembic migration file
def upgrade():
    with op.batch_alter_table("registrable_configs") as batch:
        batch.alter_column("team_reg_fee",   type_=sa.Numeric(10, 2))
        batch.alter_column("player_reg_fee", type_=sa.Numeric(10, 2))
    with op.batch_alter_table("team_registrations") as batch:
        batch.alter_column("amount_paid", type_=sa.Numeric(10, 2))
    with op.batch_alter_table("player_registrations") as batch:
        batch.alter_column("amount_paid", type_=sa.Numeric(10, 2))
```

**Add unique constraints**

```python
op.create_unique_constraint("uq_team_reg_event",    "team_registrations",  ["team", "event"])
op.create_unique_constraint("uq_team_reg_league",   "team_registrations",  ["team", "league_id"])
op.create_unique_constraint("uq_player_reg_event",  "player_registrations",["player", "event"])
op.create_unique_constraint("uq_player_reg_league", "player_registrations",["player", "league_id"])
op.create_unique_constraint("uq_headref",           "headrefs",            ["player", "event"])
op.create_unique_constraint("uq_match_name_event",  "matches",             ["name", "event"])
op.create_unique_constraint("uq_tag_name_event",    "tags",                ["name", "event"])
op.create_unique_constraint("uq_field_name_event",  "fields",              ["name", "event"])
op.create_unique_constraint("uq_sidecomp_result",   "sidecompresults",     ["comp", "player"])
```

**Add mutual exclusivity CHECKs**

Use `batch_alter_table` to add the CHECK to `team_registrations`,
`player_registrations`, and `tos`.

**Add `Player.email` and `Team.email` unique constraints**

Run the duplicate-check query first. If duplicates exist, decide policy (keep newest,
merge accounts) before applying.

**Add performance indexes**

```python
op.create_index("ix_matches_event",         "matches",              ["event"])
op.create_index("ix_points_match",          "points",               ["match"])
op.create_index("ix_match_notes_match",     "match_notes",          ["match"])
op.create_index("ix_team_reg_event",        "team_registrations",   ["event"])
op.create_index("ix_player_reg_event_pl",   "player_registrations", ["event", "player"])
op.create_index("ix_headrefs_player_event", "headrefs",             ["player", "event"])
op.create_index("ix_tags_event",            "tags",                 ["event"])
op.create_index("ix_fields_event",          "fields",               ["event"])
```

**Documentation checklist**
- [ ] Each new model class (`HeadRefAllowList`, `MatchReferee`, `MatchPlayer`, `FieldCamera`, `MatchCameraStreamStart`, `CameraTimepoint`) ships with the full docstrings
- [ ] Every Alembic migration file generated in this phase has a module-level docstring
- [ ] If the `RegistrableConfig`, `TeamRegistration`, or `PlayerRegistration` Python models are modified to change the column types, update their class docstrings to note that fee columns store exact decimal values.
- [ ] Any new Rust structs added to `frontend/src/types.rs` or `api.rs` to represent the new tables must include `///` doc comments.

**Rollback:** Every step is a `DROP TABLE`, `DROP INDEX`, or
`batch_alter_table` in reverse. Fully reversible.

CoolCatMona:
Additive changes for db schema update

Six new normalised model classes (app/models/normalised.py), one per join table the additive migration creates. Each ships with the full docstring describing what it normalises, why the unique constraint matters, and any non-obvious invariant (slot ordering, side mutual exclusivity, sequence preservation).

Monetary columns are exact decimals. RegistrableConfig.team_reg_fee, RegistrableConfig.player_reg_fee, TeamRegistration.amount_paid, PlayerRegistration.amount_paid are now Numeric(10, 2). Verified end-to-end — values round-trip through the ORM as Decimal('12.34'), not float.

Email uniqueness on Player.email and Team.email (both unique=True, index=True). Multiple NULLs allowed (SQL-standard semantics) — confirmed with a dedicated test.

Mutual-exclusivity CHECK on TeamRegistration, PlayerRegistration, and TO, identical to the one Tournament and PenaltyType already have. Rejects rows with both event and league_id set, and rows with neither set.

Alembic migration 0002_phase1_additive (migrations/versions/0002_phase1_additive.py):
100+ line module docstring covering what, why, deliberately-deferred items, pre-conditions, and rollback path.
6 op.create_table for the new tables.

Combined batch_alter_table per affected table so SQLite only rebuilds each one once.

Unique constraints implemented as op.create_index(..., unique=True) — direct SQL on SQLite, no table rebuild.
8 performance indexes per the design doc's hot-path list.

Real downgrade() implementation that reverses every operation. Verified it works end-to-end against a stamped baseline DB.

### DIFF
--- a/app/models/normalised.py
+++ b/app/models/normalised.py
@@ -1,7 +1,7 @@
 """Normalised join tables that replace previously-encoded list/blob columns.
 
-These six models normalise columns that historically stored multiple values
-in a single cell (comma-separated strings or JSON arrays). Each model has a
+These models normalise columns that historically stored multiple values in
+a single cell (comma-separated strings or JSON arrays). Each model has a
 unique constraint that enforces invariants the application code previously
 had to maintain in Python, plus real foreign keys so deletes cascade
 correctly and orphan rows cannot exist.
@@ -12,6 +12,7 @@ that happens the new tables are populated by a backfill script and kept in
 sync by dual-write code (none of which exists yet — that is later work).
 The schema definitions here are intentionally additive; nothing on the
 existing schema is modified.
+
 """
 
 from __future__ import annotations
@@ -20,7 +21,6 @@ from app.domain.enums import WinnerSide
 from app.models.base import db
 from app.models.constants import (
     LONG_NAME_LEN,
-    LONG_URL_LEN,
     URL_SLUG_LEN,
     USER_ID_LEN,
     UUID_LEN,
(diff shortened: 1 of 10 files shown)


app/models/normalised.py
```diff
@@ -0,0 +1,228 @@
...
+        db.String(USER_ID_LEN),
+        db.ForeignKey("players.id"),
+        nullable=False,
+    )
+    side = db.Column(db.Enum(WinnerSide), nullable=False)
+
+    __table_args__ = (db.UniqueConstraint("match_uuid", "player_id", name="uq_match_players_match_player"),)
+
+
+class FieldCamera(db.Model):  # type: ignore[misc]
```
reid23: this could probably be merged with the Camera table. This was for live-streamed fields while the Camera table is for recorded fields post editing, but they're both displayed in the same place.

lwk its been a while since i built this part and tbh the bandwidth required to livestream is so high that I don't think its a feature anyone will use in the forseeable future. It might be best to just get rid of this

plus this is mildly confusing since this is not the same as the list of cameras from the recording page, which is tracked through the filesystem (:skull: i know)

app/models/user.py
```diff
@@ -40,7 +43,7 @@ class Player(UserMixin, db.Model):
     name = db.Column(db.String(SHORT_NAME_LEN), nullable=False)
     pw_hash = db.Column(db.String(AUTH_STRING_LEN), nullable=True)  # Nullable for Google OAuth users
     google_id = db.Column(db.String(AUTH_STRING_LEN), unique=True, nullable=True)  # Google OAuth ID
-    email = db.Column(db.String(AUTH_STRING_LEN), nullable=True)  # Email from Google
+    email = db.Column(db.String(AUTH_STRING_LEN), unique=True, index=True, nullable=True)
```
reid23: email is currently just for people to add contact info - it's not actually used by the login process, google or otherwise. 

teams should be able to have the same email because they may be from the same club (ie, mistcrow and ursae would have the same contact info)

app/models/user.py
```diff
@@ -99,7 +105,7 @@ class Team(UserMixin, db.Model):
     pw_hash = db.Column(db.String(AUTH_STRING_LEN), nullable=True)  # Nullable for Google OAuth users
     google_id = db.Column(db.String(AUTH_STRING_LEN), unique=True, nullable=True)  # Google OAuth ID
     phone = db.Column(db.String(PHONE_LEN))
-    email = db.Column(db.String(AUTH_STRING_LEN), nullable=True)  # Updated to match Player, can be from Google
+    email = db.Column(db.String(AUTH_STRING_LEN), unique=True, index=True, nullable=True)
```
reid23: same as with the comment on Player emails - they're not used for login so they shouldn't be unique

app/models/user.py
CoolCatMona: OH, okay

I definitely saw the google oath and im like, "yeah this is probably unique"; ill fix this!

app/models/user.py
CoolCatMona: Greatly appreciate the note here, will fix

reid23 (approved):
sounds good. no notes really. makes sense why it should be structured this way

PR #146, opened 2026-04-25
Dev

reid23:
(i should really come up with a regular schedule for when these merges happen)

but this is mainly to get the alembic snapshot of production before any further architectural changes are made


CoolCatMona:
Is there anything you want me to look at specifically? This is just the changes from `dev`, right?

In the existing database there are some orphan  foreign keys from some cancelled team registrations that may appear as warnings

This shouldnt cause anything strange to happen unless we were to write `UPDATES` for these deleted teams, for whatever reason... (Sorry Hailey4Prez2028 and JesperMalleneaux)

pyproject.toml
```diff
@@ -19,12 +19,70 @@ dependencies = [
...
+    "unit: Unit tests",
+    "integration: Integration tests",
+    "slow: Slow running tests",
+]
+
+[tool.mypy]
+python_version = "3.12"
+strict = true
+
+[tool.ruff]
```
CoolCatMona, line 65: fun fact: this is what dogs say

reid23:
yeah no real thoughts this is just to get the snapshot in before any other changes are made

PR #156, opened 2026-05-06
[Feature] Enable Organizer Quick Register for streamlining registration process for day events

CoolCatMona:
Adds a tournament-organizer-driven registration flow that lets TOs add existing players and existing teams to a tournament with auto-confirmed   
  registrations. 
                                                                                                                                               
  This subsumes and replaces the public-facing "Day Event Sign-In" feature that was previously on this branch (anonymous account creation,     
  walk-up self-signin).
## User-facing behavior                                                                                                                         
    - Teams - search for a team, select, optionally edit the per-tournament pseudonym (defaults to the team's account name), submit.           
  - Both tabs share a "Checked in this session" log so a TO can see what they just did across many sequential check-ins.                       
  - Successful check-ins clear the form and refocus the search input for the next entry.                                                       
                                                     

<img width="1024" height="370" alt="image" src="https://github.com/user-attachments/assets/fcd12058-1c59-4a41-8eff-7f68837062e0" />

<img width="1075" height="588" alt="image" src="https://github.com/user-attachments/assets/30e73bef-a2c0-4380-8149-a1afd5bdeb99" />

<img width="1087" height="887" alt="image" src="https://github.com/user-attachments/assets/4e903d08-e4dc-401a-8339-53ac8441256d" />

                                                                                          
## Authorization model                                                                                                                          
                                                                                                                                               
  - Frontend Admin panel button is gated on organizer_checkin_enabled && is_current_user_to(...).                                              
  - GET /_api/<url>/checkin-info and the two POST endpoints both enforce TO authorization at the service layer (returning HTTP 403 to non-TO
  authenticated users; 401 from the existing _api/* unauthorized_handler for anonymous callers).                                               
  - Player and team registrations created via this flow are CONFIRMED, paid (paid=True, amount_paid=0), and skip the regular
  team_registration_open / player_registration_open window checks - this is the TO-override path.                                              
                                                            
### Data                                                                                                                     
                                                            
  - Existing CANCELLED/REJECTED PlayerRegistration rows for the same (tournament, player) are reused (flipped to CONFIRMED) instead of creating
   duplicates.
  - Existing CANCELLED TeamRegistration rows for the same (tournament, team) are reused (TeamRegistrationStatus only has CONFIRMED and         
  CANCELLED - no REJECTED).                                                                                                                    
  - The n_max_teams cap is honored for organizer-driven team registration; the error message includes the current count and limit (e.g.
  "Maximum teams reached (3/3)").                                                                                                              
  - Pseudonym validation runs the existing team_pseudonym_char_error validator (rejects , and ::).
                                                                                                                                               
## What changed                                              
                                                                                                                                               
### Backend                                                   

  - New column Tournament.organizer_checkin_enabled: Boolean NOT NULL DEFAULT False in app/models/tournament.py.                               
  - New exception OrganizerCheckinDisabledError in app/exceptions.py.
  - New service methods in app/services/registration_service.py:                                                                               
    - organizer_checkin(tournament_url, *, actor_user_id, actor_user_type, player_id, team_id, jersey_number, jersey_name, 
  waiver_legal_name_signature) -> Result[PlayerRegistration, ArctosError]                                                                      
    - organizer_register_team(tournament_url, *, actor_user_id, actor_user_type, team_id, pseudonym) -> Result[TeamRegistration, ArctosError]
  - New routes in app/routes/registration.py:                                                                                                  
    - POST /_api/<url>/checkin                                                                                                                 
    - GET  /_api/<url>/checkin-info                                                                                                            
    - POST /_api/<url>/checkin-team                                                                                                            
  - Modified app/routes/tournaments.py:                                                                                                        
    - POST /_api/create-tournament reads organizer_checkin_enabled from the form and passes it to the new Tournament.
    - POST /_api/<url>/update-settings reads the same field and persists it.                                                                   
  - Modified app/routes/_api.py: tournament-detail JSON response exposes organizer_checkin_enabled (replacing the previous is_day_event key    
  from the in-progress branch work).                                                                                                           
                                                                                                                                               
 ### Frontend (Dioxus + Rust)                                                                                                                     
                                                            
  - New page frontend/src/pages/organizer_checkin.rs rendering the OrganizerCheckin component at the route /:url/checkin. Players and Teams    
  tabs share a session log (modeled as a SessionLogEntry::{Player, Team} enum).
  - New types in frontend/src/types.rs: CheckinInfoResponse, CheckinResponse, CheckinTeamResponse, CheckinTeam. Renamed Tournament.is_day_event
   -> organizer_checkin_enabled.                                                                                                               
  - New API client functions in frontend/src/api.rs: checkin_info, checkin, checkin_team. Reuses the existing players_list and teams_list for
  search.                                                                                                                                      
  - New checkboxes on the Create Tournament, Add Event to League, and Tournament Settings forms (new_tournament.rs, league_new_tournament.rs,
  tournament_settings.rs). API client create_tournament(...) gains a new organizer_checkin_enabled: bool parameter.                            
  - Tournament-home Admin panel button labeled "Event Check-in", gated on the flag + TO check.
  - Layout route-title helper updated.                                                                                                         
                                                                                                                                               
## Tests                                                                                                                                        
                                                                                                                                               
  - New tests/test_organizer_checkin.py covering service and route layers for both player and team flows (~25 tests). Cases include flag-off,  
  non-TO auth rejection, happy paths, unknown player/team, duplicate-confirmed rejection, CANCELLED-row reuse, n_max cap with count-in-message,
   waiver required, JSON content-type, missing required fields.                                                                                
                                                            
## Documentation

  - README bullets added for the new endpoints (app/routes/README.md) and service methods (app/services/README.md). Model docstring updated    
  (app/models/tournament.py).
                                                                                                                                               
## High-level system impacts                                 

  - No backwards-compatibility concern. The day-of sign-in endpoints (POST /day-signin, GET /day-signin-info) and the Tournament.is_day_event  
  JSON key existed only on this feature branch - they were never merged to dev or main. Nothing in production references them.
  - Authorization is service-layer enforced. Both POST routes call into RegistrationService which performs the TO check. Direct calls to the   
  service (e.g. from scripts) cannot bypass authorization.                                                                                     
  - No CORS / origin changes, no new third-party dependencies, no changes to authentication.
  - Performance is comparable to existing registration endpoints; both new POST routes are O(1) plus a single DB commit.                       
                                                                                                                                               
## Migrations                                                                                                                                   

None


reid23 (commented):
merge conflicts now, plus a few thoughts:

imo the wording should be changed so it's more clear that this lets TOs register players. if they're registered already, they don't need to check in. to me, a "checkin" process entails confirming to the organizers that you're actually there, that you've paid, picking up spirit of jugger bands, etc. - things that are necessarily day-of. 

currently the checkin process usually involves checking the "manage registration" page and ensuring that people are paid. could we unify that, this functionality, and side comp signups? (or do we even need side comp signups? can that be something automatically assigned for everyone registered for the tournament?)

CoolCatMona:
I can fold in the check-in process to the manage registration, makes sense not to have that be duplicated.

The motivation for this would be for something like a league day, it's inconvenient for some folks to sign up to a league day in advanced, and this could help with folks who show up day-of the event. 

Additionally, this would bypass registration fees, which are not needed for these kinds of events from what I understand

I think that side competitions should still be separate, as a player I don't always compete in every side competition, so populating that list with every player registered seems a bit extraneous imho

CoolCatMona:
Updated

TO's can now "Quick Register" players, which has the same functionality as the day event check-in

This appears in the "Manage Registration" tab

image

PR #157, opened 2026-05-07
[Feature] Create Side Competitions

CoolCatMona:
(review after #156 )
adds side competition creation to ARCTOS.

This appears as a tab next to the "Tournament Information" per event
A Tournament Organizer is able to create/edit/delete side competitions
Side competitions default as `Closed` and do not allow for registration, a TO needs to open the registration.
A Tournament Organizer is able to check-in players to a side competition (**Enabling "refs" to do this is future work**)
A player who is registered for an event may check themselves into a side competition
Players registered for a side competition are assigned a unique number for that side competition

## Backend Changes

- **Models**
  - New `SideCompRegistration` table with `entry_number` (unique per comp, 1-indexed).
  - New `SideCompType` enum in `app/domain/enums.py` (DUELING, CHAIN_BREAKING, OTHER).
  - `SideComp` extended with `created_at`, `description` (Text, nullable), `registration_open` (Boolean, default False).
  - `SideComp.type` migrated from free-form string to `db.Enum(SideCompType)`.
- **Migrations**
  - `0006_sidecomp_registrations.py` - new table + `created_at` column.
  - `0007_sidecomp_open_close_description.py` - `description` and `registration_open` columns.
  - `0008_sidecomp_entry_number.py` - `entry_number` column with backfill via `ROW_NUMBER() OVER (PARTITION BY comp ORDER BY id)`.
- **Service layer (`app/services/sidecomp_service.py`)**
  - `create`, `update` (now accept `description` and `registration_open`).
  - `delete`, `list_for_event`, `get_with_registrants`.
  - `register_player` - rejects when `registration_open` is False (`RegistrationClosedError`).
  - `deregister_player`.
  - `organizer_check_in` - works regardless of `registration_open` (TO override).
  - `organizer_remove`.
  - `cancel_player_registrations_in_event` cascade helper.
  - `_next_entry_number(comp_id)` helper - assigns sequential numbers, never reuses.
- **Cascade wiring** - all event-scoped paths that flip `PlayerRegistration` to CANCELLED hard-delete matching `SideCompRegistration` rows:
  - `RegistrationService.deregister_player`
  - `RegistrationService.deregister_team` (per-player loop)
  - `app/routes/registration.py::deregister_any_player`
  - `app/routes/registration.py::deregister_any_team`
  - `app/routes/tournaments.py::delete_tournament`
- **Routes (`app/routes/sidecomps.py`)** - 10 endpoints under `/_api`:
  - Public: `GET /<url>/sidecomps`, `GET /sidecomps/<id>`.
  - TO CRUD: `POST /<url>/sidecomps`, `PATCH /sidecomps/<id>`, `DELETE /sidecomps/<id>`.
  - Player: `POST /sidecomps/<id>/register`, `POST /sidecomps/<id>/deregister`.
  - TO check-in: `POST /sidecomps/<id>/checkin`, `POST /sidecomps/<id>/uncheckin`.
  - Picker: `GET /sidecomps/<id>/eligible-players` (TO-only).
- **Detail response** includes viewer-context flags - `viewer_is_to`, `viewer_can_register` (logged-in Player + CONFIRMED event reg + comp open + not already in comp), `viewer_is_registered_in_comp` - so the frontend can gate buttons without leaking identity.

## Frontend

- **5 new pages**: `SideCompsList`, `SideCompNew`, `SideCompEdit`, `SideCompDetail`, `SideCompCheckin`.
- **`SideCompNew` / `SideCompEdit`** - name, type select, description textarea; Edit also has a registration-open switch and a Delete button.
- **`SideCompDetail`** - status badge (Open/Closed), description, viewer-conditional Edit / Check in players (TO only) / Register me / Deregister me buttons; registrants table with `#` (entry number), Player, Registered, Source columns.
- **Tournament home** - "Tournament Information" card converted into a tabbed card with "Tournament Information" + "Side Competitions" tabs.
- **Side Competitions tab** - rows are click-to-view (drop the View button); TO sees a `+` icon to add and per-row pencil/trash icons (delete confirms via `window.confirm`).
- **Standalone `SideCompsList` page** matches the same click-to-view list-group pattern.
- Types and `reqwest` API wrappers cover all routes.
- `cargo check --target wasm32-unknown-unknown` clean.

## Migrations

`make db-current` -> should be 0005
`make db-backup`
`make db-migrate` -> applies migrations for 0006, 0007, 0008
`make db-current` -> Should be 0008

## Note

This feature **only** enables side competition creation, there is no feature for actually running the side competition as of yet. I'm not sure what that would look like, but I am envisioning something to just create a score field per entrant and then the ability to add one to the score should suffice.

image
image
image
image
image
image
image


tests/test_sidecomps.py
```diff
@@ -0,0 +1,1112 @@
+"""Tests for side competition models, services, and routes."""
```
CoolCatMona, line 1: A third of this PR is tests :3

app/services/sidecomp_service.py
```diff
@@ -0,0 +1,493 @@
...
+    @staticmethod
+    def _get_tournament(tournament_url: str) -> Result["Tournament", ArctosError]:
+        from app.exceptions import TournamentNotFoundError
+        from models import Tournament
+
+        tournament = Tournament.query.filter_by(url=tournament_url).first()
+        return option(tournament).ok_or(TournamentNotFoundError(tournament_url))
+
+    @staticmethod
+    def _next_entry_number(comp_id: int) -> int:
```
reid23, line 56: - i think that players should have the same entry number across all side comps they enter in. I don't think this would significantly change the code, and it would significantly reduce confusion and the odds of misreported results
- it might be nice to add the ability to skip certain numbers (in particular, the teens can sound confusing - 15 sounds like 50, for example)

reid23 (approved):
some small thoughts; the second bullet about number skipping is less important and we can add that later but imo at the thing about keeping entry numbers consistent for players across all side comps should be done before we merge into main. 

it doesn't require any schema changes so i think this is still good to merge into dev so testing can happen there and we can get other things merged. all the functionality present here is indeed functional so i say we get it live and adjust later.

PR #158, opened 2026-05-07
[Bugfix] Fix bugs that may cause errors, remove some dead code

### LINKED ISSUE #123: Add missing imports, remove unused imports
## Missing Imports

Missing imports may cause runtime errors within the running application

- `render_template` in `app/routes/matches.py`
- `urlencode` imported from `flask_login`, not `urllib.parse` in `app/filters.py`

## Unused Imports

Unused imports make the code more difficult to read
 
Unused `Flask` import in `generate_permission_key.py`

CoolCatMona:
Bug fixes (Review after #157)

### `calculate_stream_timestamp` always returns `None` (dead body)
- The function early-returns `None` on the empty-input guard, then the file defines the camera-access-key helpers (`generate_camera_key`, `validate_camera_key`, `get_camera_key_from_request`, `require_camera_key`). The actual timestamp-calculation body (the `try:` block parsing point and stream timestamps) is indented inside `require_camera_key` *after* its `return (True, None)` — making it unreachable. Callers therefore always receive `None`. Stream-relative seek times in the UI silently fail.

### `MatchGraph.get_dependencies` / `get_dependents` reference a non-existent attribute
-  Both methods do `self.nodes_by_name.get(...)`, but the graph stores nodes in `self.nodes_by_key`. Any caller hits `AttributeError`.

### `app/routes/matches.py` uses `render_template` without importing it
- Top-level `from flask import Blueprint, request, jsonify` does not include `render_template`. `make_response` is imported only inside one inner function.
-  Routes that hit any of those lines `NameError` at runtime. Either the routes are never exercised (then they are dead and should be deleted) or the test suite does not cover them.

### Duplicate Flask route registrations
- Each pair registers the same `(path, method)` on the same blueprint. Flask keeps the last-registered view function for any overlapping `(path, method)` tuple, so one of each pair is dead code that drifts.
-  ambiguous behaviour for `/server-time` and `/stones`.

### JSON error handler returns HTTP 200 for every API error
- For requests that accept JSON, the handler ignores `e.status_code` and always returns 200 ("Keep prior behavior"). The whole point of `ArctosError.status_code` is defeated; SPAs cannot distinguish 401/403/404/400/500 by status code, only by parsing the JSON body.

### `db.create_all()` swallows all errors at startup
- Three `except Exception: pass` blocks at app boot mean schema problems, lock-pragma failures, and YouTube-resume errors are invisible. `db.create_all()` cannot ALTER existing tables, so combined with no migration discipline the schema silently drifts from the models.

### `n_max_teams` / `n_max_players` cap enforcement has a TOCTOU race

- Three near-identical "count CONFIRMED rows → compare to cap -> insert" sequences. With no row lock or unique constraint, two simultaneous registrations can both pass the cap and exceed it.

### `_next_entry_number` race on concurrent side-comp registration

- `MAX(entry_number) + 1` with no row lock. Two simultaneous registrants compute the same value; the unique constraint `uq_sidecomp_registrations_comp_entry_number` then fails one with an `IntegrityError`.

## Migrations

`make db-backup`
`make db-migrate`


app/utils/camera_helpers.py
```diff
@@ -181,6 +181,50 @@ def calculate_stream_timestamp(point_stamp, stream_start_time):
     if not point_stamp or not stream_start_time:
         return None
 
+    try:
```
CoolCatMona, line 184: moved from `require_camera_key`, might have been misplaced

app/utils/MatchGraph.py
```diff
@@ -256,16 +256,6 @@ def add_dependency(
...
-    def get_dependencies(self, match_name: str) -> Set["Dependency"]:
-        """Get the set of Dependency wrappers that the given match depends on."""
-        node = self.nodes_by_name.get(match_name)
-        return node.dependencies.copy() if node else set()
-
-    def get_dependents(self, match_name: str) -> Set[MatchGraphNode]:
-        """Get the set of nodes that depend on the given match."""
-        node = self.nodes_by_name.get(match_name)
-        return node.dependents.copy() if node else set()
-
```
CoolCatMona, line 268: Unused code

app/routes/matches.py
```diff
@@ -2,7 +2,7 @@
 Match operation routes (start, run, finalize, view).
 """
 
-from flask import Blueprint, request, jsonify
+from flask import Blueprint, request, jsonify, render_template, make_response
```
CoolCatMona, line 13: add missing imports

app/routes/matches.py
```diff
@@ -802,21 +800,21 @@ def get_selection_notes(tournament_url):
...
     event_urls = match_event_urls_for_penalties(tournament)
 
     match = Match.query.get(match_id)
     if not match or match.event not in event_urls:
-        return json_error("Match not found")
+        return json_error("Match not found", status_code=404)
 
     if not can_head_ref_match(tournament_url, current_user.id, match=match):
-        return json_error("bruh ur not a head ref")
+        return json_error("bruh ur not a head ref", status_code=403)
```
CoolCatMona, line 825: its forbidden.

app/routes/_api.py
```diff
@@ -3913,81 +3913,6 @@ def _injury_json(inj):
...
-        if not custom_date:
-            injury.stamp = None
-        else:
-            try:
-                injury.stamp = datetime.strptime(custom_date, "%Y-%m-%d")
-            except ValueError:
-                return jsonify({"error": "Invalid date format. Use YYYY-MM-DD."}), 400
-
-    db.session.commit()
-    return jsonify(_injury_json(injury))
```
CoolCatMona, line 3991: There was already CRUD implemented for injuries, multi-method handlers like these were also unused

app/routes/matches.py
```diff
@@ -1372,73 +1370,6 @@ def complete_match(tournament_url):
...
-
-    # Filter files based on user permissions
-    # Only show "Classic" and "Snare" unless user is in the allowed list
-    user_can_see_all = current_user.is_authenticated and current_user.id in ALLOWED_USERS
-
-    if not user_can_see_all:
-        # Filter to only show "Classic" and "Snare" (case-insensitive)
-        mp3_files = [f for f in mp3_files if f["display_name"].lower() in ["classic", "snare"]]
-
-    return render_template("stones_player.html", mp3_files=mp3_files)
```
CoolCatMona, line 1435: Same route as `stones_list`

app/routes/matches.py
```diff
@@ -1372,73 +1370,6 @@ def complete_match(tournament_url):
...
-def server_time():
-    """Return current server time in unix timestamp format."""
-    import time
-
-    return jsonify(
-        {
-            "server_time": time.time(),
-            "timestamp": datetime.now(timezone.utc).isoformat(),
-        }
-    )
```
CoolCatMona, line 1448: Duplicate code / blueprint

Already exists in `_api.py`

app/utils/responses.py
```diff
@@ -30,15 +28,15 @@ def json_success(data: Optional[Dict[str, Any]] = None, status_code: int = 200)
...
 
-    Defaults to HTTP 200 to preserve historical API compatibility.  Pass an
-    explicit *status_code* when the client needs a real error code.
+    Defaults to HTTP 400 (Bad Request). Pass an explicit *status_code* when the
+    route should signal a different error code (e.g. 403, 404).
 
     Args:
         message: Human-readable error description.
-        status_code: HTTP status code to use (default 200).
+        status_code: HTTP status code to use (default 400).
```
CoolCatMona, line 39: 200 is the OK status code. Just changes to `Bad Request` by default

app/__init__.py
```diff
@@ -191,14 +194,13 @@ def set_sqlite_pragmas(dbapi_connection, _connection_record):
                         cur.close()
 
     except Exception:
-        pass
+        logger.exception("Failed to register SQLite pragma listener")
```
CoolCatMona, line 215: Still bare exceptions, but now give context instead of failing silently

In the future, we should be able to raise specific exceptions when we encounter these

app/services/registration_service.py
```diff
@@ -124,33 +132,44 @@ def register_team(tournament_url: str, team_id: str, pseudonym: str) -> Result["
...
-
-        team_registration = TeamRegistration(event=tournament_url, team=team_id, pseudonym=pseudonym)
 
-        # Auto-mark as paid if registration fee is zero
-        from app.utils.helpers import get_registrable_config
+        team_registration = TeamRegistration(
+            event=tournament_url,
+            team=team_id,
+            pseudonym=pseudonym,
+            status=TeamRegistrationStatus.PENDING,
```
CoolCatMona: Initially marked as PENDING, Once the team is added and the team does not exceed the cap, will be marked as CONFIRMED

app/services/sidecomp_service.py
```diff
@@ -67,6 +67,43 @@ def _next_entry_number(comp_id: int) -> int:
...
+            try:
+                db.session.commit()
+                return reg
+            except IntegrityError as exc:
+                db.session.rollback()
+                last_exc = exc
+                continue
+        assert last_exc is not None
+        raise last_exc
+
```
CoolCatMona, line 106: Prevents race conditions if two players are added at the same time

reid23 (approved):
omg omg omg yippee

(i don't see any issues and this was long overdue)

PR #159, opened 2026-05-07
[Documentation] scheduling algorithm docs and others

reid23:
added documentation for the scheduling algorithm as it currently stands. taken from an old issue i wrote a long time ago. 
Also added documentation for all the environment variables used for configuration and the ASS cheat sheet from the help modal.

### DIFF
--- a/docs/env-vars.md
+++ b/docs/env-vars.md
@@ -1,0 +1,92 @@
+# Environment Variables
+
+This is a list of all environment variables that can be used to
+configure Arctos.
+
+`SECRET_KEY`
+: the flask secret. generate randomly.
+
+`ARCTOS_CORS_DEV`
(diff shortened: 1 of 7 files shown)


docs/env-vars.md
```diff
@@ -0,0 +1,92 @@
+# Environment Variables
```
CoolCatMona, line 1: Do you think the codebase might benefit from an `.env.example`?

docs/side-comps.md
```diff
@@ -0,0 +1,7 @@
+# Side Comps
```
CoolCatMona, line 1: : )

docs/env-vars.md
reid23: added!

PR #160, opened 2026-05-07
[Bugfix] Backend Helpers

### LINKED ISSUE #124: HTTP 200 returned for error responses
`app/utils/responses.py` and `app/error_handlers.py` document intentional HTTP 200 responses for many error conditions.

This is acknowledged as historical behavior, but it makes it harder for API clients, reverse proxies, and monitoring tools to distinguish success from failure using standard HTTP semantics. 

New endpoints should use appropriate 4xx/5xx status codes.

CoolCatMona:
### TO permission checks
- **New decorators** in `app/utils/decorators.py`:
  - `@require_tournament_organizer(message)` (existing, made JSON-aware via the new `wants_json(request)` helper)
  - `@require_league_organizer(message)` (new)
  - Both delegate to a shared private `_require_organizer` factory.
- **`PermissionService.is_league_organizer(league_url, user)`** added.
- **`PermissionService.is_tournament_organizer`** now consults `league_id` for tournaments attached to a league
- **Deleted `is_not_TO()`** legacy helper from `tournaments.py`.
- **Migrated** every inline `TO.query.filter_by(...).first()` permission check across the route layer (`tournaments.py`, `_api.py`, `registration.py`, `sidecomps.py`) and service layer (`registration_service.py`, `sidecomp_service.py`). Routes that take the URL slug from form/query data use `PermissionService.is_*_organizer(...)` directly; routes with the slug in the path use the decorator.

### Result -> JSON
- Migrated every hand-written `match res: case Ok(...): jsonify(...) case Err(err): jsonify({"error": ...}), 400` block to `json_from_result(res, ok_to_payload=...)` from `app.utils.result_helpers`.
- Deleted `_err_response` from `routes/sidecomps.py`.
- **Standardizes on `json_success` shape**: success responses now flat-merge `success: true` alongside the payload. See "Behavior changes worth knowing" below.

### User-type discriminator
- New **`UserType`** StrEnum in `app/domain/enums.py`: `PLAYER = "player"`, `TEAM = "team"`. StrEnum equality preserves backward compat with raw-string DB columns.
- `PermissionService.user_type(user)` now returns `Option[UserType]` instead of `Option[str]`.
- New **`current_user_type()`** ergonomic helper in `app/services/_common.py` for callers that need the bare string for SQL filters or serialization.
- Migrated every `current_user.__class__.__name__.lower()` site.

### Tournament lookup
- New **`get_tournament_or_err(tournament_url) -> Result[Tournament, ArctosError]`** in `app/services/_common.py` (404 on miss, otherwise Ok).
- Migrated every per-service ad-hoc lookup (`match_service`, `match_start_eligibility`, `schedule_import_export_service`, `sidecomp_service`, `registration_service`).
- Existing private `_get_tournament` helpers in `RegistrationService` and `SideCompService` are now thin delegations to the public helper.

### Naive UTC
- New **`now_utc_naive()`** in `app/utils/datetime_helpers.py`.
- Migrated 35 call sites. SQLAlchemy column defaults dropped their `lambda: ...` wrappers - the helper itself is callable, so `default=now_utc_naive` (no parens) is sufficient.

### JSON body guard
- New **`@require_json_body()`** decorator in `app/utils/decorators.py`. Returns 415 for non-JSON Content-Type, 400 for malformed JSON, otherwise stashes the parsed body on `flask.g.json_body` (defaulting to `{}` for empty body).
- Migrated 10 of 12 sites. The 2 skipped sites (`_api.py`) are inside dual-method (GET+POST) routes where the JSON requirement is conditional on `request.method == "POST"` - the decorator runs unconditionally and would 415 valid GET requests. Documented in commit message.

### Plumbing helpers also added
- **`wants_json(request)`** extracted from the duplicated logic in `error_handlers.py`. Used by all three new/updated decorators.
- **`resolve_actor(user_id, user_type) -> Player | Team | None`** in `_common.py` for service-layer code that takes the actor as id+type strings and needs a user object to call PermissionService.

## Bug fixes

**Latent: league-season TOs were denied access on league-attached tournaments.**
`PermissionService.is_tournament_organizer` previously always queried `TO.event=tournament_url`. For tournaments attached to a league (`Tournament.league_id IS NOT NULL`), the TO records live at the league level, not per-event - so a legitimate league organizer would get a 403 trying to use TO-only buttons on any tournament under their league.

The fix mirrors the existing `to_entries_for_tournament` logic from `registration_resolver.py`: load the tournament; if `league_id` is set, query by `league_id`; otherwise by `event=url`. Standalone tournaments (`league_id IS NULL`) are unchanged.


app/utils/datetime_helpers.py
```diff
@@ -94,3 +94,17 @@ def parse_datetime_local_to_utc(dt_string: str) -> datetime:
     # Convert to UTC and strip timezone info for storage
     utc_dt = local_dt.astimezone(timezone.utc)
     return utc_dt.replace(tzinfo=None)
+
+
+def now_utc_naive() -> datetime:
```
CoolCatMona, line 99: Replaces instances of `datetime.now(timezone.utc).replace(tzinfo=None)`

This is a one-liner anyways, but I figure it's easier to type; can easily be reverted/removed if its not an issue / over engineered.

app/services/permission_service.py
```diff
@@ -23,36 +23,41 @@ class PermissionService:
...
+        from app.domain.enums import UserType
+
         if user is None:
             return Null()
         if is_player(user):
-            return Some("player")
+            return Some(UserType.PLAYER)
         if is_team(user):
-            return Some("team")
+            return Some(UserType.TEAM)
```
CoolCatMona, line 44: use Enums rather than strings

app/domain/enums.py
```diff
@@ -62,6 +62,17 @@ class RegistrationStatus(StrEnum):
...
 
+class UserType(StrEnum):
+    """Account type for a Flask-Login user.
+
+    Members compare equal to their bare-string DB representation, which
+    keeps storage shape compatible with the legacy ``user_type`` columns.
+    """
+
+    PLAYER = "player"
+    TEAM = "team"
```
CoolCatMona, line 73: Seemed like a good candidate for enums

app/utils/decorators.py
```diff
@@ -38,19 +102,64 @@ def require_tournament_organizer(
...
+    """
+    return _require_organizer(
+        param_name="league_url",
+        check_fn=PermissionService.is_league_organizer,
+        redirect_path_fn=lambda slug: f"/leagues/{slug}",
+        message=message,
+    )
+
+
+def require_json_body():
```
CoolCatMona, line 137: Replaces

```
    if not request.is_json:
        return jsonify({"success": False, "error": "Content-Type must be application/json"}), 415
    data = request.get_json() or {}
```

as a decorator.

reid23 (approved):

PR #161, opened 2026-05-08
[Bugfix] Code deduplication / simplification

CoolCatMona:
### New types / helpers

- **`Scope`** value object in `app/services/_common.py` - frozen dataclass with `Scope.event(url)` / `Scope.league(url)` factories, `is_event` / `is_league` accessors, and a `__post_init__` invariant enforcing exactly-one-URL.
- **`ScopeContext`** internal helper in `registration_service.py` bundling `(scope, parent, cfg)` after the per-scope load.
- **`PermissionService.is_tournament_organizer_of(tournament, user)`** - overload that takes a pre-loaded Tournament so hot service paths skip the redundant fetch. Existing `is_tournament_organizer(url, user)` now delegates.

### register/deregister unification

8 service methods in `app/services/registration_service.py` collapsed to 4. Public surface is now:

```python
register_team(scope, team_id, pseudonym)
register_player(scope, player_id, team_id, ...)
deregister_team(scope, team_id)
deregister_player(scope, player_id)
```

Routes wrap the URL: `Scope.event(tournament_url)` (4 sites in `routes/registration.py`) or `Scope.league(league.url)` (4 sites in `routes/_api.py`). Internal bodies branch on `scope.is_league` only where event/league behavior actually differs (open-check, fee column, row construction, sidecomp cascade event-only, league match-in-progress check across all tournaments in the league).

### League IN-clause subquery

Two sites in `registration_service.py` previously materialized `[t.url for t in Tournament.query.filter_by(league_id=...).all()]` and used the result in `Match.event.in_(...)`. Replaced with a single SQL subquery: `select(Tournament.url).where(Tournament.league_id == ...)`.

### manage_api dedup

~150 lines of duplication between `tournament_manage_api` and `league_manage_api` extracted to a private `_serialize_manage(scope, search_query, search_type, cfg)` helper. Both route handlers become 10-15 line wrappers. The `LeagueTournament` fake-class adapter is gone. New `team_registrations_for_scope` / `player_registrations_for_scope` helpers added in `registration_resolver.py` alongside the existing `*_for_tournament` functions (which other callers still use).

### N+1 query fixes

| Site | Before | After |
|---|---|---|
| `SideCompService.get_with_registrants` | per-row `Player.query.get` | 2 queries (registrations + batched players) |
| `routes/sidecomps.list_for_event` | per-comp `COUNT(*)` | 1 grouped query with `func.count` + outerjoin |
| `routes/sidecomps.eligible_players` | per-row Player + per-row TeamRegistration | 2 batched queries + dict assembly |
| `SideCompService.cancel_players_in_event` (new) | per-player loop calling `cancel_player_registrations_in_event` (~2K queries when 1K players are cancelled) | 1 comp lookup + 1 `DELETE WHERE comp IN (...) AND player IN (...)` (2 queries total) |
| `tournament_detail` | per-team `Team.query.get` + per-team `player_registrations_for_tournament(team_id=...)` + per-row Player + per-row Player/Team for TOs | 2 batched fetches + `collections.Counter` for team counts + split-by-user-type batched fetches for TOs |
| `league_detail` | same patterns as `tournament_detail` (mirror route) | same fix shape, with the resolver migrated to `_for_scope` variants where available |
| `team_profile` | `Tournament.query.all()` (R7) + per-player `Player.query.get` (R6c) + site-wide `MatchNote.query.filter(...)` then Python-side filter (R8) | filtered Tournament query + batched Player fetch + SQL JOIN pushing the team1/team2 filter into the database |
| `player_profile` | per-note `Match.query.get` (twice, in two loops) + per-reg `Tournament.query.filter_by` + per-reg `League.query.get` + per-reg `TeamRegistration.query.filter_by` for pseudonyms | 5 batched fetches + dict assembly; `_team_pseudonym` closure simplified to dict access |

All field shapes preserved exactly. The batched-fetch + dict-assembly idiom is uniform across the four detail/profile routes.

### Other

-  `app/routes/matches.py` had one straggler `match res:` block. Migrated to `json_from_result`.
-  27 sites of `current_user.__class__.__name__ == "Player"|"Team"`  replaced with `is_player(user)` / `is_team(user)`. One ternary at `_api.py` migrated to `current_user_type()`. `is_player`/`is_team` promoted to module-level imports in `_api.py`
- 2 hot service paths in `registration_service.py` migrated from `is_tournament_organizer(url, user)` to `is_tournament_organizer_of(tournament, user)` since the Tournament was already loaded. Saves one DB fetch per call. The `sidecomp_service._require_to` site keeps the URL-based call (no Tournament loaded).

## Migrations

None


app/routes/notes.py
```diff
@@ -216,7 +217,7 @@ def get_point_notes(tournament_url: str):
 
     # Check if user is a head ref (for full access to all notes)
     is_head_ref = False
-    if current_user.is_authenticated and current_user.__class__.__name__ == "Player":
+    if current_user.is_authenticated and is_player(current_user):
```
CoolCatMona, line 227: Convenience function to replace old `.__class__.name__`

app/services/_common.py
```diff
@@ -79,3 +81,58 @@ def current_user_type() -> str:
                 "current_user_type() called without an authenticated Player/Team; "
                 "is @login_required missing?"
             )
+
+
+@dataclass(frozen=True)
+class Scope:
+    """Identifies a registration / manage context as either an event or league.
```
CoolCatMona, line 85: Previously, we had duplicate methods for leagues and tournaments.
Bodies differed very slightly depending on if something was a tournament or a league.

This is a scope that flags the entity as an `event` or `league`, which can be used in deduplicated methods.

app/services/registration_resolver.py
```diff
@@ -11,6 +11,83 @@
...
+
+    Args:
+        scope: A :class:`~app.services._common.Scope` identifying event or league.
+        status: Filter by this :class:`~app.domain.enums.TeamRegistrationStatus` (default
+            ``CONFIRMED``). Ignored when *exclude_cancelled* is ``True``.
+        exclude_cancelled: When ``True``, return all statuses except ``CANCELLED``.
+
+    Returns:
+        List of :class:`~app.models.registration.TeamRegistration` records.
+    """
```
CoolCatMona, line 39: Rather than having separate methods to get registrations, we can have a single method with the `scope` keyword argument.

app/services/registration_resolver.py
```diff
@@ -11,6 +11,83 @@
...
+        scope: A :class:`~app.services._common.Scope` identifying event or league.
+        team_id: When supplied, limit results to players registered for this team.
+            Ignored when *unattached_only* is ``True``.
+        unattached_only: When ``True``, return only players without a team.
+        statuses: List of :class:`~app.domain.enums.RegistrationStatus` values to include.
+            Defaults to ``[PENDING_TEAM_APPROVAL, CONFIRMED]``.
+
+    Returns:
+        List of :class:`~app.models.registration.PlayerRegistration` records.
+    """
```
CoolCatMona, line 68: Same as above

app/services/sidecomp_service.py
```diff
@@ -205,10 +205,11 @@ def get_with_registrants(comp_id: int) -> Result[tuple, ArctosError]:
...
         )
-        registrants = []
-        for r in rows:
-            p = Player.query.get(r.player)
-            registrants.append((r, p))
+        player_ids = [r.player for r in rows]
+        players_by_id = {
+            p.id: p for p in Player.query.filter(Player.id.in_(player_ids)).all()
+        } if player_ids else {}
+        registrants = [(r, players_by_id.get(r.player)) for r in rows]
```
CoolCatMona, line 210: Example of an **N+1** query, this will run a query for r in rows(), which, as you can imagine, gets expensive the more rows we have.

The fix would be to use just two queries, the first to get `ids`, the second to run a batch query, then the data structure is just assembled in pure python afterwards.

reid23 (approved):

PR #163, opened 2026-05-08
[bugfix] table creation on init

reid23:
this way the app doesn't create empty tables when we add new things, which is good since we should have all db structure changes be explicit. If we don't have a table that we expect, the app should not start.

fixes the issue where alembic migrations fail due to new tables already existing because the app was started.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -171,13 +171,6 @@ def create_app(config: dict | None = None) -> Flask:
 
     except Exception:
         pass
-    # Ensure tables exist (safe to call on startup)
-    try:
-        with app.app_context():
-            db.create_all()
-    except Exception:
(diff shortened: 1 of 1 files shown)


app/__init__.py
CoolCatMona, line 180: Yeah, my fault, I probably should have removed this myself.

I think it might be a good idea to include a `make db-backup` and `make db-migrate` as part of a deploy script in the future, which would essentially remove the manual step that we're taking now of running these migrations manually.

CoolCatMona (approved):

app/__init__.py
reid23: yeah ill add to the deploy now. i should probably put the deploy script in git and make it less sketchy

PR #166, opened 2026-05-08
[bugfix] match name uniqueness constraint

reid23:
updated `0002_phase1_additive` to enforce a different uniqueness constraint which allows duplicate match names for BREAK and JOIN match types on different fields.


CoolCatMona:
For the second failing test, if we are allowed to have duplicate names for STATIC / SAFE / FAST, we need to update the constraints, otherwise we can just wrap that block with a `pytest.raises(IntegrityError)`

CoolCatMona:
For test 1, I think the older fields are still being referenced, `uq_matches_name_event` -> `unique_with_field` and `unique_without_field`

CoolCatMona (approved):
Since you've already applied 0002 (and might have applied 0003), you would need to downgrade to a previous migration, if the downgrade proves to be impossible, you can just use the backup database that was (hopefully) created as a baseline

PR #176, opened 2026-05-13
[Feature] Shortnames

### LINKED ISSUE #33: Shortnames
(Including the `bug` tag bc I think the fact that you can have a super long team name that messes with layouts counts as a bug)

## Objectives
- make schedules and brackets more readable by abbreviating long team names
- make match descriptions more concise so that future potential notifications are usable
- limit the length of all names in a sane way (players too)

## Key Results
- add an optional "shortname" for teams to add upon registration with a small fixed max length
- add a "short" filter which tries to use the shortname and falls back to a truncated version of the full name 
- add length limits to places in templates that currently are unbounded, especially those corresponding to strings that may be very long

## Implementation Plan
Probably wait until after #28 is done since then we'll know the actual limits

CoolCatMona:
Adds an optional per-event team shortname (max 12 chars) captured at
registration time. Long pseudonyms are no longer forced into space-constrained
UI (schedule cells, bracket lines, match cards) - the frontend displays the
shortname when present and falls back to a 12-character ellipsis truncation
when it isn't.

### Storage

- New nullable `shortname VARCHAR(12)` column on `team_registrations`.
- New constant `SHORTNAME_LEN = 12` in `app/models/constants.py`.
- Single Alembic revision `0008_team_registration_shortname`; nullable,
  no backfill, downgradable. Existing rows have `NULL` and fall back to
  truncation until users edit their registration.

### Service layer

- `RegistrationService.register_team` and `register_team_as_to` accept
  an optional `shortname` kwarg. Empty / whitespace-only is normalised
  to `None` so the column never holds an empty string.

### HTTP routes

- `POST /<tournament_url>/register-team`, `POST /_api/leagues/<.../register-team`,
  and `POST /<tournament_url>/register-team-as-to` read shortname from
  the request body and pass to the service.
- `PUT /_api/leagues/<.../registrations/team/me` and
  `PUT /_api/tournaments/<.../registrations/team/me` implement a
  three-way semantic for the field: key absent -> column unchanged;
  key present and null/empty/whitespace -> column cleared to NULL;
  key present and non-empty -> column set to trimmed value.

### Output payloads

- Every JSON serializer that emits `pseudonym` now also emits
  `shortname` (or `team1_shortname` / `team2_shortname` / `team_shortname`
  for match-shaped payloads). Covers ~20 sites across `_api.py`,
  `matches.py`, `sidecomps.py`, `tournaments.py`, and
  `team_stats_service.py`.

### Rust frontend

- `frontend/src/types.rs`: ~14 structs gained
  `shortname: Option<String>` with `#[serde(default)]` so older
  payloads still deserialize cleanly.
  `UpdateTeamRegistrationRequest.shortname` additionally uses
  `skip_serializing_if = "Option::is_none"` so callers that don't set
  the field don't accidentally clear the column via the three-way
  semantic.
- New `frontend/src/display.rs`: tiny `short_or_truncate(full,
  shortname) -> String` helper. Uses `chars().count()` so Unicode
  works correctly. Six co-located unit tests.
- Registration form input added on four surfaces:
  `tournament_register.rs`, `league_register.rs`, the
  `edit_registration_modal.rs` (Dioxus signal-bound), and `manage.rs`'s
  register-team-as-to flow. All use `maxlength="12"` and helper text.
- `short_or_truncate` applied at: bracket node labels
  (`bracket.rs`), schedule table cells and timeline chips
  (`schedule.rs`), match-card lists and conflict-modal radio labels
  (`match_page.rs`). NOT applied to autocomplete dropdowns or
  main match-detail headers.
- The schedule highlight-search filter compares against the raw
  (non-truncated) pseudonym so typing a full team name still matches
  rows whose displayed label was abbreviated.

## Migration

Single Alembic revision `0008_team_registration_shortname`

## Screenshots

Adding Shortnames
image

Default shortname in schedule view
image

Bracket View
image
(Note: sample database doesnt have images i promise this is normal : ))

## Other

I think some things got merged without passing through the linter, so some of the changes from that are here as well.

### DIFF
--- a/tests/unit/test_string_length_validation.py
+++ b/tests/unit/test_string_length_validation.py
@@ -25,54 +25,55 @@ def test_player_name_rejects_value_longer_than_short_name_len(test_db):
 
 @pytest.mark.unit
 def test_player_name_accepts_value_exactly_at_short_name_len(test_db):
-    """A 100-char name (exactly SHORT_NAME_LEN) is accepted."""
+    """A 100-char name (exactly SHORT_NAME_LEN) is accepted.
+
+    Guards against an off-by-one regression that changes ``len(value) > N``
+    to ``>=`` in the validator.
+    """
     player = Player(id="lenok1", pw_hash="dummy")
-    player.name = "x" * 100  # should not raise
-    assert player.name == "x" * 100
-
-
-@pytest.mark.unit
-def test_player_name_accepts_value_shorter_than_short_name_len(test_db):
-    """A short name is accepted unchanged."""
-    player = Player(id="lenok2", pw_hash="dummy")
-    player.name = "Alice"
-    assert player.name == "Alice"
+    player.name = "x" * 100  # must not raise ValidationError
 
 
 @pytest.mark.unit
 def test_nullable_string_accepts_none(test_db):
-    """Assigning None to a nullable String column does not raise."""
+    """Assigning None to a nullable String column does not raise.
+
+    Exercises the ``isinstance(value, str)`` guard: without it, ``len(None)``
+    would raise ``TypeError`` from inside the validator.
+    """
     player = Player(id="lennone", name="N", pw_hash="dummy")
-    player.location = None  # Player.location is nullable String(SHORT_NAME_LEN)
-    assert player.location is None
+    player.location = None  # must not raise ValidationError or TypeError
 
 
 @pytest.mark.unit
 def test_validator_passes_through_non_string_values(test_db):
-    """The length validator only checks strings; other types are not its concern."""
+    """Non-string values bypass the validator entirely.
+
+    Exercises the ``isinstance(value, str)`` guard for non-None, non-str
+    inputs. SQLAlchemy's own type handling owns the error path for type
+    misuse; the length validator is not it.
+    """
     player = Player(id="lenint", name="N", pw_hash="dummy")
-    # Assigning an int to a String column is a type misuse, but the length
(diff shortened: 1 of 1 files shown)


app/routes/_api.py
```diff
@@ -2857,7 +2870,8 @@ def _team_pseudonym_and_photo(tournament, team_id):
         pseudonym = team.name
     if not pseudonym:
         pseudonym = team_id
-    return pseudonym, profile_photo
```
CoolCatMona, line 2838: Worth noting that some of these functions that originally return a tuple are now polyamorous and return a thruple.

Appropriate changes should exist in the calling functions, but its worth noting when/if these functions are used.

setup/apt-packages.txt
```diff
@@ -1,2 +1,3 @@
+build-essential
```
CoolCatMona: I setup an environment on a windows machine with a fresh wsl install and needed to add this; can totally make this a separate PR if you want

app/models/constants.py
```diff
@@ -51,6 +51,11 @@
 """Tag name strings used in ASS (Arctos Schedule Script) expressions.
 Kept shorter than display names to stay practical as identifiers."""
 
+SHORTNAME_LEN: Final[int] = 12
+"""Optional short team display alias used in space-constrained UI
+(schedule cells, bracket lines, match cards). Stored on
+``TeamRegistration.shortname``."""
+
```
CoolCatMona: 12 was picked kind of arbitrarily

PR #179, opened 2026-05-14
event check-in and side comps

reid23:
still need to test side comps on dev


reid23:
i should really stop merging things without doing in-depth testing but i discovered some more problems with side comps:
- any player can create a new side comp from the `/<url>/sidecomps` page, regardless of their registration or TO status in the tournament
- players cant register themselves? it seems like only TOs can?

and then these arent bugs but just thoughts:

- do we really need the `/<url>/sidecomps` page? what does it do that the sidecomps tab on the tournament page doesn't do?
- the "session log" on the quick register page feels weird from a ui perspective - maybe just have a little badge next to people's names that says `registered` and shows their number or something, and hide the 'quick register' button? 
- kinda related to the above but TOs should def be able to see a player's number when they quick register them for a side comp

CoolCatMona:
I didn't encounter the players being able to create a sidecomp, so I will look into that
Players should be able to register themselves, only if they're registered for the league/tournament

I think overloading the sidecomps tab is not very user friendly, but maybe I'm misunderstanding what you're asking. I can start drawing mockups in the future though, and happy to refactor

So like if I type "Jordan" and register him, and then type "Jordan" again, it shows a badge next to his name?
- I feel like as a ref/to, having a million players come up and ask "am I registered?" Is something that I've seen and experienced, so having a quick reference was the purpose of that.
TOs seeing a players number is definitely something I can add that wouldn't take too long I think.

reid23:
aight maybe these words will be better written

### bugs

i re-tested this now that its not midnight and heres what i found:

- i think i made a mistake when testing bc non-TOs definitely can't register. They can see the create sidecomps button on the sidecomps page (not tab) but that's not an actual bug.
- Players can only register themselves for side comps on standalone tournaments - when I tried to register for a side comp on a league event, the button doesn't exist. 
- TOs can't register players on league events either - there's an error saying "player is not registered"

### sidecomps page

i dont mean getting rid of `/<url>/sidecomps/<id:int>` - these pages for each comp are good and useful. 
I just mean getting rid of the `/<url>/sidecomps` page, which just has the same list of sidecomps as the tab on the tournament page.
This change would require:
- getting rid of the page
- changing the `/<url>/sidecomps/new` form to redirect back to the tournament homepage (with the sidecomp tab open) instead of the sidecomp page
- making the "delete side comp" button redirect back to the tournament homepage w/ sidecomp tab open instead o the sidecomp page (or just getting rid of the button, since we already have delete functionality on the tab on the tournament page)
- changing the "back to side comps" link to instead just redirect back to the tournament homepage w/ sidecomp tab open

i dont think the `sidecomps` page adds anything, since the edit, delete, and new sidecomp functionality is already present on the tab on the tournament page.

### Quick Register UI

this was more of a passing thought; just that instead of throwing the registered players in the "session log", we could just convert the "quick register" button to a thing that says registered and shows their number. 

Especially since that way, if a player is already registered, the TO gets positive confirmation that the player is registered when they search their name (instead of just coming up with nothing)

PR #182, opened 2026-05-15
[Feature] API Refactor Prep, add facades, surface urls 

### LINKED ISSUE #174: Refactor `_api.py` monolith
## Problem

This issue started as the plan to split the old `_api.py` monolith. That specific file has since been split, but the underlying developer-experience problem is not fully finished: several route modules still mix HTTP handling, authorization checks, serialization, database mutation, scheduling logic, and workflow orchestration in the same file.

The result is a route layer that is easier to navigate than before, but still hard for a new developer to modify confidently. It is not always obvious whether new code belongs in `app/routes/`, `app/services/`, `app/serializers/`, or `app/utils/`.

## Current cleanup targets

Large or workflow-heavy route areas include:

- `app/routes/matches.py`
- `app/routes/registration.py`
- `app/routes/leagues.py`
- `app/routes/tournaments/matches_admin.py`
- `app/routes/tournaments/read.py`
- `app/routes/tournaments/recordings.py`
- `app/routes/tournaments/scheduling.py`

Related boundary files to inspect:

- `app/routes/README.md`
- `app/services/README.md`
- `app/serializers/README.md`
- `app/services/`
- `app/serializers/`
- `frontend/src/api.rs`

## Suggested approach

1. Treat this as a follow-up cleanup issue, not as a request to resurrect `_api.py`.
2. Pick one route area at a time and define the target boundary before editing.
3. Keep routes focused on request parsing, authorization/decorator use, service calls, and response shaping.
4. Move reusable workflow logic into `app/services/`.
5. Move response construction that is shared across endpoints into `app/serializers/`.
6. Preserve the existing URL surface unless a separate product issue explicitly changes it.
7. Update the route/service/serializer READMEs whenever the boundary becomes clearer.

## Definition of done

- The old `_api.py` split is reflected in the issue text and docs.
- At least one large route module is made meaningfully easier to scan without changing its public URL behavior.
- New contributors can tell from the docs where to put a route, service function, serializer, or utility.
- Tests cover any moved behavior.

CoolCatMona:
Pure-additive setup for the upcoming routes refactor.

Five new modules and one new helper, currently **pure re-exports** of
private helpers that live in `app/routes/_api.py`. **No behavior change**.

| New module | Re-exports |
|---|---|
| `app/serializers/user_serializer.py` | `_user_json` -> `user_json` |
| `app/serializers/registration_serializer.py` | `_player_reg_waiver_api` -> `player_reg_waiver_api`; `_serialize_manage` -> `serialize_manage` |
| `app/serializers/league_serializer.py` | `_require_league` -> `require_league`; `_league_to_dict` -> `league_to_dict` |
| `app/serializers/tournament_serializer.py` | `_tournament_to_dict` -> `tournament_to_dict`; `_team_name_for_match` -> `team_name_for_match` |
| `app/utils/profile_photo_helpers.py` | `_profile_photo_upload_dir` -> `profile_photo_upload_dir`; `_safe_profile_photo_filename` -> `safe_profile_photo_filename` |
| `app/utils/datetime_helpers.py` (modified) | `_dt_iso` -> `dt_iso` (added alongside existing helpers) |

Each facade module includes its module docstring and a corresponding
`.rst` toctree entry under `docs/api/`.

Adds a single regression test that locks the full `(methods, rule)`
set of `app.url_map` against a compact text fixture. The test
deliberately ignores endpoint names because blueprint-prefixed names
will shift as routes move (`_api.leagues_list` ->
`leagues.leagues_list`, etc.). `HEAD` and `OPTIONS` are excluded
because Flask auto-generates them.

| File | Purpose |
|---|---|
| `tests/test_url_surface.py` | The test itself. Marked `@pytest.mark.unit`. |
| `tests/fixtures/url_surface.txt` | 198 lines, one rule per line as ``METHODS RULE`` (e.g. ``GET,POST /_api/<tournament_url>/start-match``). The current URL surface. |
| `app/routes/README.md` | One new bullet pointing future contributors at the regen recipe. |

When a future portions of the refactor genuinely changes the URL surface (adds a new route,
changes a method, etc.), the fixture is regenerated as part of that
PR.

The fixture and the test exist specifically to guard the in-flight
`_api.py` refactor; the final cleanup removes both, since the routes will all live in their permanent
homes by then


tests/test_url_surface.py
```diff
@@ -0,0 +1,51 @@
+r"""URL surface snapshot test.
```
CoolCatMona, line 1: This file will be removed once api refactor is done : )

tests/fixtures/url_surface.txt
```diff
@@ -0,0 +1,198 @@
+DELETE /_api/<tournament_url>/penalty-types/<int:pt_id>
```
CoolCatMona, line 1: This file will be removed when api refactor is done : )

reid23 (approved):

PR #184, opened 2026-05-19
[bugfix] stricter url validation requirements

reid23:
im too lazy to write an issue but a user tried to register a tournament with a url that started with a `/`, which caused a bunch of problems. 
probably should update the explanation text to not have an example that starts with a slash, but the validation definitely needs to happen either way.


CoolCatMona (approved):
Seems pretty straightforward

PR #185, opened 2026-05-19
remove `EXTERNAL_BASE_URL`

### LINKED ISSUE #152: add arctos.usnja.org domain
**Is your feature request related to a problem? Please describe.**
now that this is an option for a free domain id like to add it. it'd be easier to tell people and easier to remember and all that.

**Describe the solution you'd like**
remove all uses of EXTERNAL_BASE_URL and instead get the base from the request. then add the new domain to all the relevant google apis. 
then update the nginx config to serve arctos on both domains.

**Describe alternatives you've considered**
could use a cloudflare worker to accomplish this but that costs a little bit of money ($60/yr) so id rather not

reid23:
this way it's all based on the path of the request, which should allow us to host simultaneously on both `events.californiajugger.org` (for backwards compatability) and `arctos.usnja.org`.

also updated the footer

closes #152

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,7 +7,6 @@ S3_ENDPOINT_URL=https://s3.us-east-005.backblazeb2.com
 AWS_REGION=us-east-005
 AWS_ACCESS_KEY_ID="[25 char censored string]"
 AWS_SECRET_ACCESS_KEY="[31 char censored string]"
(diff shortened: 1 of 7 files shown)


CoolCatMona:
Can you rebase onto dev to remove the duplicate commits

CoolCatMona (approved):

PR #186, opened 2026-05-20
Fix league side-comp registration

CoolCatMona:
Follow-up to #179.

This fixes side-comp registration for league-backed tournaments. The previous lookup only treated a player as registered when their PlayerRegistration event matched the side-comp event directly, which breaks league-scoped registrations. The side-comp service now resolves eligibility through the tournament event and league scope, so player self-registration, TO registration, and the detail-page viewer_can_register flag all use the same rule.

System impact: limited to side-comp eligibility and registration checks. No migration needed.

Tested:
tests/test_sidecomps.py::test_register_player_succeeds_with_league_scoped_registration tests/test_sidecomps.py::test_register_player_as_to_succeeds_with_league_scoped_registration tests/test_sidecomps.py::test_route_detail_viewer_can_register_with_league_scoped_registration tests/test_sidecomps.py::test_route_eligible_players_returns_league_scoped_registrations

### DIFF
--- a/app/routes/sidecomps.py
+++ b/app/routes/sidecomps.py
@@ -50,18 +50,11 @@ def _detail_payload(sc, registrants):
     viewer_can_register = False
     viewer_is_registered_in_comp = False
     if current_user.is_authenticated:
-        from app.domain.enums import RegistrationStatus
-        from models import PlayerRegistration
-
         viewer_is_to = PermissionService.is_tournament_organizer(sc.event, current_user)
         if is_player(current_user):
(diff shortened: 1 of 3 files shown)


CoolCatMona:
I added a frontend screenshot for this branch.

It uses mocked API data so the page can be captured without a local seeded database. The bit to check is that an eligible league-registered player sees `Register me`; the backend behavior is covered by the pytest cases in the PR body.

Side-comp detail with Register me button

reid23 (approved):

PR #187, opened 2026-05-20
Show quick-register status in search results

CoolCatMona:
Follow-up to #179.

This changes the TO quick-register flow so search results still show players who are already in the side comp. Those rows now show a registered badge and entry number, including immediately after a successful register action, so the page no longer needs a separate session log.

System impact: the eligible-players response now includes sidecomp_registered and entry_number. Existing consumers can ignore those fields. No migration needed.

Tested:
tests/test_sidecomps.py::test_route_eligible_players_marks_sidecomp_registered_players tests/test_sidecomps.py::test_route_eligible_players_returns_league_scoped_registrations

### DIFF
--- a/app/routes/sidecomps.py
+++ b/app/routes/sidecomps.py
@@ -247,7 +247,7 @@ def deregister_player_as_to(comp_id: int):
 @bp.route("/sidecomps/<int:comp_id>/eligible-players", methods=["GET"])
 @login_required
 def eligible_players(comp_id: int):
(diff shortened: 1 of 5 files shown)


CoolCatMona:
I added a screenshot for the quick-register flow.

This was captured after clicking `Quick Register` for Alex. Both Alex and Mina stay in the result list, and both rows show their side-comp entry numbers.

Quick-register results showing registered entry numbers

reid23 (approved):

Issue #199, opened 2026-05-28
Fix Schedule Editing


reid23:
## Problem

Schedule editing needs to be safer and more understandable. Right now, changing a tournament schedule can affect match order, dependencies, refs, tags, start times, breaks, joins, and ASS expressions. Some of those invariants are implicit in the code rather than documented as a clear editing model.

The main product goal is to let a tournament organizer edit matches without needing to be a developer and without accidentally invalidating the running schedule.

## Developer context

Useful files to inspect:

- `frontend/src/pages/schedule.rs`
- `frontend/src/components/ass_entry.rs`
- `app/routes/tournaments/matches_admin.py`
- `app/routes/tournaments/scheduling.py`
- `app/services/schedule_import_export_service.py`
- `app/services/dual_write.py`
- `app/services/match_start_eligibility.py`
- `app/utils/scheduling.py`
- `app/utils/parser.py`
- `app/utils/dependencies.py`
- `docs/arctos-schedule-script.md`
- `docs/scheduling.md`

## Things this should handle

- Circular dependencies.
- Obvious double-booking.
- Conditional or outcome-dependent double-booking risk.
- Missing teams, registrations, matches, fields, tags, or refs.
- ASS expressions that do not parse or do not resolve.
- Renaming matches and keeping all references/tags in sync.
- Editing `_initial` fields while preserving the resolved team/ref fields correctly.
- Inserting, moving, or deleting matches without corrupting the schedule order.
- Dynamic matches that can be safely constrained without a previous match.
- Editing once a tournament has already started.
- TOML import/export, or a deliberate removal/deprecation of that path.
- Static breaks at the start of a day.
- Named anchors that can be completed by TOs to advance multiple schedule points together.

## Suggested approach

This is too large for one implementation PR. Split it into smaller issues or PRs around stable invariants:

1. Write down the schedule invariants that edits must preserve.
2. Add validation that can report schedule problems without requiring the organizer to understand internal data structures.
3. Fix reference/tag propagation for match renames and `_initial` field changes.
4. Decide whether the linked-list ordering model should stay or be replaced by rebuilding order from queries and saving the result.
5. Handle running-tournament edits explicitly instead of treating them as the same as pre-event edits.
6. Add regression tests for every invariant that gets changed.

## Definition of done

- A developer can read the docs/tests and understand what makes a schedule valid.
- The edit UI reports actionable problems before or immediately after an unsafe edit.
- Match references, tags, refs, breaks, and ordering remain consistent after edits.
[...]

PR #205, opened 2026-05-29
[Feature] ASS entry box component

### LINKED ISSUE #200: ASS Entry UI fixing
## Objectives
- entering ASS code is very tricky right now. it should be easy to type and think about.

## Key Results
- when you type an open square bracket (or when the innermost braces are square or when the closest open bracket on the left is curly), autofill to team options, and show as token
- when you type an open curly bracket (or when the innermost braces are curly or when the closest open bracket on the left is curly), autofill to match options, and show as token
- when you type any open bracket, auto-close it.
- when you type an open paren, autofill to valid functions & their documentation (current is incomplete)
- improve error messages & static analysis

reid23:
## Summary

adds a component for ASS entry with suggestions, completion, and validation that actually make sense.  
also helps split up the massive Thing that is `schedule.rs`

i know nobody uses this feature!! i promise its relevant to do this issue first; its so that we can migrate the team1/team2/refs entry boxes to ASS as a part of the editing refactor. this autocomplete should make it easier not to fuck up when entering data.

## What changed

- returns errors/warnings that actually make sense (ie, `unexpected '(', line 1, col 32`)
- checks that braces are balanced
- difflib fuzzy search for suggestions
- distinguish between nonexistant tags and unset tags 🫪
- check for invalid references
- display more match info (teams, field) in the autocomplete when in curly brackets
- updated DSL_FUNCTIONS const
- make the ass-related frontend stuff into a component and move it into a separate file in the components directory

still todo: 
- [x] ensure it can work with different box sizes
- [x] make it a bit more compact for use in team and ref entries
- [x] make it actually check the expression when i stop typing for long enough (rn its just when i focus away from the box)
- [ ] ~~get rid of those awful chains of four `if let`s~~ nope my rust-analyzer is cooked and gives errors about the current year being before 2024 when this was stable and i dont care enough to fix it

## Migration impact

none! the migration will come when we actually start using this for team1/team2/refs.

## Test plan

- [x] `make test` passes locally
- [x] Manual verification - tested expressions with:
  - tags (that exist, don't exist, or exist but are not set)
  - team literals which do or dont exist
  - team literals in the `matchname::winner` format
  - match literals which do or do not exist
  - conditionals
  - lambdas, cons, cat, cdr, and map (over numbers)

## Linked issues

Closes #200 

## Screenshots / repro

turns out that putting tokens inline with editable text is rlly hard and you need to use a canvas or something and do all of the ui custom. im not about that, so this will be good enough

image
image
image

---

### Pre-review checklist

- [x] `make lint` and `make format` are clean.
- [ ] ~~If a new module was added under `app/`: corresponding `app/<area>/README.md` and `docs/api/*.rst` are updated (per `CLAUDE.md`).~~
- [x] If developer workflow changes: `README.md`, `CONTRIBUTING.md`, or `TESTING.md` updated.

### DIFF
--- a/app/routes/tournaments/scheduling.py
+++ b/app/routes/tournaments/scheduling.py
@@ -831,15 +831,18 @@ def validate_dsl(tournament_url):
         SymbolicTeam,
         SymbolicMatch,
         Lambda,
+        Nil,
+        _format_dsl_value,
         _infer_types,
     )
 
     def serialize_value(value):
         """Convert the interpreted value to a JSON-serializable format."""
-        if isinstance(value, (int, bool, type(None))):
+        if isinstance(value, Nil):
+            return None
+        if isinstance(value, (int, bool)):
             return value
         elif isinstance(value, list):
-            # Recursively serialize list elements
             return [serialize_value(item) for item in value]
         elif isinstance(value, Team):
             # Return team ID
@@ -860,33 +863,6 @@ def validate_dsl(tournament_url):
             # Fallback to string representation
             return str(value)
 
-    def value_to_string(value):
-        """Convert the interpreted value to a readable string representation."""
-        if isinstance(value, (int, bool, type(None))):
-            return str(value)
-        elif isinstance(value, list):
-            # Format as Lisp-like expression
-            if len(value) > 0 and isinstance(value[0], str):
-                # Preserved expression - format as s-expression
-                return "(" + " ".join(value_to_string(item) for item in value) + ")"
-            else:
-                # Data list
-                return "[" + ", ".join(value_to_string(item) for item in value) + "]"
-        elif isinstance(value, Team):
-            return f"[{value.obj.id}]"
-        elif isinstance(value, Match):
-            return f"{{{value.obj.name}}}"
-        elif isinstance(value, SymbolicTeam):
-            return f"[{value.literal}]"
-        elif isinstance(value, SymbolicMatch):
-            return f"{{{value.literal}}}"
-        elif isinstance(value, Lambda):
-            # Lambda objects shouldn't appear in final results, but handle gracefully
-            params_str = " ".join(value.params) if value.params else ""
-            return f"(lambda ({params_str}) ...)"
-        else:
-            return str(value)
-
     data = request.get_json()
     expression = data.get("expression", "").strip()
 
@@ -905,7 +881,7 @@ def validate_dsl(tournament_url):
         # the simplified form (with chips). Suppressing on equality hid useful previews
         # for expressions whose literal text already happens to be in canonical form
         # (e.g. anything containing an unset [tag::Foo] that stays symbolic).
-        simplified = value_to_string(result)
+        simplified = _format_dsl_value(result)
 
         # Treat unresolvable team/match references as errors so the user notices typos.
         if warnings:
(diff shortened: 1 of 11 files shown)


CoolCatMona:
You can't tell your opps what this branch's name is

reid23:
one of these days my digital footprint will catch up to me

thankfully that's not today!!

CoolCatMona:
How does it feel knowing that I am probably the only other person that looks at these commit messages

CoolCatMona:
2c: I think that exposing domain specific languages to the **user** are not very user friendly. Have you considered making this like a form/interactable object that the user can leverage?

reid23:
> How does it feel knowing that I am probably the only other person that looks at these commit messages

i feel grateful. both that you're reading them and that nobody else will

reid23:
> 2c: I think that exposing domain specific languages to the **user** are not very user friendly. Have you considered making this like a form/interactable object that the user can leverage?

yes, the reasoning was:
- I want this to remain generic, so I don't have to implement special-case functionality for every specific type of bracket that people want to use. it feels like every tournament the bracketing system changes, and i hate bracketing, so I don't want to deal with it myself. I'd rather give people something fairly expressive and let them figure it out.
- I have no hope of designing a form that can acheive anything the user might want to do, and if I did manage to pull it off, it'd probably be super complicated and i'd have a whole bunch of options that are weirdly named and confusing.
- Using an existing [text] language would be convenient for users, but it would open up a whole bunch of security vulnerabilities from a language I don't understand. Plus, if there's tons of language features, it makes it harder to process and propagate schedule promises through. I don't want to analyze someone's python code and try to figure out what matches it depends on.
- Using a visual language (like scratch, ev3/nxt, blender graphs, simulink, etc), would be a HUGE effort to implement, and i wouldn't even know where to start on language design

In retrospect, however, the reason this exists is that I realized Arctos had no way to do things like double elim brackets where some games may not be played, so I rushed to implement the most fun thing I could think of to solve the problem without *really* changing the architecture of the system. and thus the skip condition, and by extension ASS, was born. I liked how if you didn't *need* to use it, you didn't have to learn anything about it, so I figured it'd be fine.

This is probably a really good discussion to have soon; as part of #199 I was planning to make the team and ref entries ASS expressions in an effort to phase out the `MatchName::winner` syntax in favor of `(winner {MatchName})`. that would allow team entries to benefit from the same rich analysis i've implemented here (if we just use the component in this PR for match/ref entries). (both still work since in ass `MatchName::winner` is a valid team literal, but if we just document that less, we can eventually remove it and just do a migration). 
Do you think that's a good idea? I think it's really nice since it allows teams to be specified in a way more expressive way than we have now, and extensions to the language could add new features without much effort.

one example of something I've been thinking about: people want a way to say like "pod 1 rank 2" and have it automatically adjust so the TOs dont have to panic and look through results and assign tags halfway through the day. We could have a "event variables" table where people can define custom expressions, and then people could define all the intermediate results they want and have teams auto-update.

realistically this is all pointless tho because nobody is ever going to use anything this complicated, at least in the next year, and perhaps it would be prudent to only develop features once theres actual demand for them.

tldr im just a silly little emacs user and i love ~~ass~~ my little dsl but beyond that i dont really know what to do. what do you think?

reid23:
created [this issue](https://github.com/reid23/arctos/issues/215#issue-4615455970) to continue this discussion. i think this pr is fine to merge either way though since it still helps

PR #213, opened 2026-06-03
[Refactor] Delete _api.py and finalize the route refactor

CoolCatMona:
## Summary

Deletes `app/routes/_api.py` and ties off the `_api.py` monolith refactor. After the nine route-move PRs landed, `_api.py` held no routes at all - only the `_dt_iso` and `_check_to` helpers (plus some dead leftover function bodies). This PR relocates those two helpers to permanent homes, removes the module and its blueprint registration, and retires the URL-surface drift gate that guarded the whole effort.

## What changed

- **Deletes `app/routes/_api.py`** and drops its `register_blueprint(_api_bp)` wiring in `create_app`.
- **`dt_iso`** in `app/utils/datetime_helpers.py` gets its real implementation, replacing the lazy re-export of `_api._dt_iso`.
- **`serialize_manage`** in `registration_serializer.py` now imports `tournament_to_dict` and `dt_iso` from their real homes (`tournament_serializer` / `datetime_helpers`) instead of the lazy `_api` import - whose `_tournament_to_dict` target had already moved out in the previous PR, so this also fixes a latent import break on the event-scope manage path.
- **`_check_to`** becomes `check_tournament_organizer` in `app/utils/decorators.py` - a boolean companion to `require_tournament_organizer`. `content.py` and `penalty_types.py` call it from there.
- **Removes the URL-surface drift gate** (`tests/test_url_surface.py` + `tests/fixtures/url_surface.txt`). It was introduced to guard the refactor stack against accidental URL drift; with `_api.py` gone and every route in its permanent home, it has served its purpose.
- **`app/routes/README.md`**: drops the "fall back to `_api.py`" wording and the url-surface-fixture step.

## Test plan

- [x] `just unit` passes locally - 231 passed (was 232; the removed URL-surface snapshot test accounts for the -1)
- [x] `just integration` passes locally - 60 passed
- [x] App boots with no `_api` blueprint; `app.url_map` still has the same 197 rules (the surface is unchanged because `_api.py` had no routes left)

The URL surface is unchanged, so the existing integration tests continue to cover every endpoint by URL.

Note: `just lint` reports one pre-existing F401 in `app/models/validators.py` (`URL_SLUG_LEN` unused) - it's on `dev` already and unrelated to this PR, so left alone.

## Linked issues

Closes #174

## Screenshots / repro

No UI or API-shape change. Same URLs, same methods, same payloads.

---

### Pre-review checklist

- [x] Title prefixed with `[Refactor]`.
- [x] Branch name follows `category/name` (`feature/refactor-finalize`).
- [x] Targeting `dev`, not `main`.
- [x] `just format` clean; `just lint` clean on touched files (one unrelated pre-existing warning noted above).
- [x] Tests updated: the URL-surface gate is intentionally removed now that the refactor is complete.
- [x] Developer workflow docs: `app/routes/README.md` updated.


reid23:
yippee yippee yippee!!!!

PR #219, opened 2026-06-10
Match edit stability, schedule warnings, and ASS entry box component.

### LINKED ISSUE #200: ASS Entry UI fixing

### LINKED ISSUE #217: Pre-MHO immediate schedule edit fixes
## Objectives
- reduce the odds of entirely fucking up the schedule by trying to edit it live
- make data entry easier by showing some common errors

## Key Results
- disallow editing match names
- prevent editing a match after it has started (status IN_PROGRESS, COMPLETED, or SKIPPED)
- add "scheduled start time" and relevant migration (just copy nominal start time)
    - this still kinda has some undefined behavior/weird edge cases but it's much better. since this doesnt change when you solve schedule. so you can't have a weird situation where every time you click "recompute schedule" it changes through some set of states in a chain
        - aside: good nerd snipe problem: prove if there exists a finite schedule for which the FSM with the state transition function "recompute schedule" never converges and never repeats itself.
    - make the time-based dependency be based on this instead of nominal start time
- when updating the _initial fields, also propagate the update to the static cache
- fix insertion into doubly linked list
- add error display and handling that makes sense. For the errors listed below, don't check for them at match creation/edit, and allow them to be fine. we should check them after adding the match, and list them in a modal accessible with a little button in the edit toolbar.
    - nonexistant teams
    - cyclic dependencies
    - matches that don't exist
    - double-booked teams (naively! doesn't need to look at all future possibilites, just what's actually written. for now.)

## Implementation Plan
going to need a migration to add the schedule start time, but that should be easy. we can just copy nominal start time from old matches.

reid23:
## Summary

primarily pre-MHO changes working towards #199. doesn't do everything, but kind of does the bare minimum to make sure MHO goes well:
- proper error handling
- better warnings about common issues
- stabilize topological sort

also includes the smart ASS entry box 

## What changed

See #217 

- creating, editing, and deleting matches now never fails (previously schedule issues would lead to opaque "server error" messages and sometimes the changes would go through anyway and sometimes they wouldn't)
    - changing match names is prohibited. yes this is kinda a bandaid fix but it is better than nothing, and deleting and re-creating a match isn't the end of the world.
- no more cyclic changes from recomputing the schedule by switching topological sort tiebreaker metric from nominal_start_time (which changes dynamically) to scheduled_start_time (which is set without knowledge of the true match start/end times)

some of the docs have been updated but not as much as I had hoped. I just really want to get this out before MHO.

see #200 
- autocomplete that works for ASS
- shows more info for matches (ie, teams, field)
- type checking for ASS
- error messages for simple ASS errors
- new component that manages all this
- removed `cons` because im not dealing with implementing that properly
- added `quote` and the standard shorthand so people can construct lists

## Migration impact

     - Alembic revisions added: 0009

very simple. copies existing info from nominal_start_time. safe to revert.

## Test plan

- [x] `just coverage-check` passes locally, or CI coverage is sufficient
- [x] Manual verification:
  - no data lost when deploying to staging
  - tested a few different schedule scenarios; worked as expected. See the dev PR.
  - tested all ass functions individually and type checking through chains that included everything

## Linked issues

Closes #217 
Closes #200 
Refs   #199 

### Pre-review checklist

- [ ] ~~Branch name follows `category/name` (`feat/...`, `bugfix/...`, `refactor/...`).~~
- [ ] ~~Targeting `dev`, not `main` (unless this is a release PR).~~
- [ ] ~~If a new module was added under `app/`: corresponding `app/<area>/README.md` and `docs/api/*.rst` are updated.~~
- [ ] ~~If developer workflow changes: `README.md`, `CONTRIBUTING.md`, or `TESTING.md` updated.~~


CoolCatMona (approved):

CoolCatMona:
looking forward to the rollout day before the tournament : )

PR #230, opened 2026-07-17
[Feature] Tunable Stones!

### LINKED ISSUE #225: Fix Stones
## objectives

- make stones actually sync
- remove headaches with wires
- make stones not sound funny sometimes

## key results

- switch to a single sync to an existing set of time servers (not the
  arctos server)
  - store in local storage to persist across reloads.
- add the ability to adjust the sync
  - one adjustment for the clock offset
  - one adjustment for the audio delay
  - button to re-ping time server
- update running matches
  - update how stone calculations start since they wont be accurate in
    server time anymore? not sure if any changes are actually needed
    but there might be
  - add docs to ref page for how to do this
  - on "run match", if it's a stones match, check how stale the config
    is. if it anything doesn't seem right, ask for manual
    recalibration of the clock offset.
- long-term: add automatic sync via mic? eww audio & signal processing
  are horrible

## implementation

- may need to be careful about people having multiple tabs open. maybe
  restrict to only one? or use the session id? should we allow people
  to have stones on one tab and ref a match on the other?

reid23:
## Summary

stones sync is kinda shitty rn. lets make it better.

## What changed

- change to ping multiple times and take the minimum to reduce the effect of network latency asymmetry
- change to make measurement variance increase with longer RTT (according to an empirically derived exponential)
- increase process variance to something roughly based on how much I saw my clock change one day over an hour. plus some safety factor
- add calibration procedure that walks the user through adjusting:
    - audio delay compensation: bc bluetooth has delay
    - total clock offset: just move the whole clock a constant amount. if theres a persistent offset.

## Test plan

- [x] `just coverage-check` passes locally, or CI coverage is sufficient
- [x] Manual verification:
  - local testing must be done with a manually edited `api.rs` to make it ping prod, so we actually get nonzero RTT. same goes for testing on dev; that server is very close since it's in berkeley, so we'll get better than expected results!

## Linked issues

Closes #225 

## Screenshots / repro

image
image
image

### Pre-review checklist

- [ ] ~~If a new module was added under `app/`: corresponding `app/<area>/README.md` and `docs/api/*.rst` are updated.~~
- [ ] ~~If developer workflow changes: `README.md`, `CONTRIBUTING.md`, or `TESTING.md` updated.~~

### DIFF
--- a/frontend/Cargo.toml
+++ b/frontend/Cargo.toml
@@ -38,8 +38,11 @@ web-sys = { version = "0.3", features = [
     "MediaDeviceInfo", "MediaStreamTrack",
     "Storage",
     "StorageManager",
+    "StorageEvent",
 ] }
 [target.'cfg(target_arch = "wasm32")'.dependencies]
+dioxus-web = "0.7"
+futures = { version = "0.3", default-features = false, features = ["std", "async-await"] }
 gloo-timers = { version = "0.3", features = ["futures"] }
 idb = "0.6"
 js-sys = "0.3"
--- a/frontend/src/pages/run_match.rs
+++ b/frontend/src/pages/run_match.rs
@@ -1,7 +1,6 @@
 use crate::Route;
 use crate::api;
 use crate::components::PenaltyDisplay;
-use crate::stones_filter::BayesianOffsetFilter;
 use dioxus::prelude::*;
 #[cfg(target_arch = "wasm32")]
 use gloo_timers::callback::Interval;
@@ -431,12 +430,13 @@ fn quick_timer_current(start_time: f64, start_stones: u32, now_server: f64) -> i
     start_stones as i64 - (now_beat - start_beat).max(0)
 }
 
-/// Stones elapsed = global 1.5s beat boundaries crossed. Uses Bayesian filter for server time when end is None (ongoing point).
+/// Stones elapsed = global 1.5s beat boundaries crossed. Uses the shared time
+/// sync for corrected server time when end is None (ongoing point).
 #[cfg(target_arch = "wasm32")]
 fn stones_elapsed_with_filter(
     stamp_opt: Option<&str>,
     end_opt: Option<&str>,
-    filter: &Signal<BayesianOffsetFilter>,
+    time_sync: &crate::time_sync::TimeSync,
 ) -> u32 {
     const BEAT: f64 = 1.5;
     let start = match stamp_opt.and_then(|s| chrono::DateTime::parse_from_rfc3339(s).ok()) {
@@ -447,12 +447,10 @@ fn stones_elapsed_with_filter(
         if let Ok(parsed) = chrono::DateTime::parse_from_rfc3339(e) {
             parsed.timestamp_millis() as f64 / 1000.0
         } else {
-            let client_time = js_sys::Date::now() / 1000.0;
-            client_time + filter.read().get_mean()
+            time_sync.server_now_secs()
         }
     } else {
-        let client_time = js_sys::Date::now() / 1000.0;
-        client_time + filter.read().get_mean()
+        time_sync.server_now_secs()
     };
     let start_beat = (start / BEAT).floor() as i64;
     let end_beat = (end / BEAT).floor() as i64;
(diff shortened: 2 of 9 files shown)


reid23:
@CoolCatMona at your earliest convenience can i get a review on this

mostly on the UI/UX side, if you had thoughts on how to make this less confusing for users
and if you think this is generally a good direction

reid23:
merge conflicts are non-trivial since the utility timer needs to be ported to this new remote timer system. will get to it probably tmr

reid23:
@CoolCatMona pleeeaaase :point_right: :point_left: can i get a revieww 

im gonna ping u on strava soon if this doesnt work

frontend/src/pages/run_match.rs
```diff
@@ -990,10 +964,12 @@ pub fn RunMatch(url: String, match_id: String) -> Element {
...
                         });
                     } else {
-                        // Start new point — optimistic: add pending point, set current_point
-                        let pending_id =
-                            format!("pending-{}", chrono::Utc::now().timestamp_millis());
-                        let stamp_iso = chrono::Utc::now().to_rfc3339();
+                        // Start new point — optimistic: add pending point, set current_point.
+                        // Anchor the point to corrected server time so it lands on the same
+                        // 1.5s grid the stone counter measures against.
+                        let point_start_ms = point_start_server_millis(&time_sync);
```
CoolCatMona: Point start now uses corrected server time, but both end paths still use raw `chrono::Utc::now()`. Any device or manual clock offset can change the persisted duration and stone count. Please use the same corrected clock for both.

frontend/src/pages/stones.rs
```diff
@@ -68,57 +86,51 @@ fn StonesPlayerWasm(stones_val: ReadSignal<Option<Result<StonesResponse, String>
...
-                    while beat_time <= max_wall && scheduled < 5 {
+                    while next_server_beat - server_now <= SCHEDULE_AHEAD_SEC && scheduled < 5 {
                         if !is_playing_sig() {
                             break;
                         }
-                        let delay = beat_time - now;
+                        let delay = next_server_beat - server_now;
                         if delay > 0.05 && delay < 60.0 {
-                            let audio_time = audio_now + delay;
+                            let audio_time = audio_now + delay + audio_delay;
```
CoolCatMona, line 134: A positive audio delay is added here, so Bluetooth compensation schedules playback later instead of earlier. Please subtract it or label the value as a signed timing correction.

frontend/Cargo.toml
```diff
@@ -38,8 +38,11 @@ web-sys = { version = "0.3", features = [
     "MediaDeviceInfo", "MediaStreamTrack",
     "Storage",
     "StorageManager",
+    "StorageEvent",
 ] }
 [target.'cfg(target_arch = "wasm32")'.dependencies]
+dioxus-web = "0.7"
```
CoolCatMona, line 44: `cargo test` uses the configured WebAssembly target and fails because `display.rs` imports `wasm_bindgen_test`, but this manifest does not declare it. Please add the test dependency and run frontend tests in CI.

frontend/src/time_sync.rs
```diff
@@ -0,0 +1,455 @@
...
+        let touches_calibration = match key.as_deref() {
+            Some(k) => k == CLOCK_OFFSET_KEY || k == AUDIO_DELAY_KEY,
+            None => true,
+        };
+        if touches_calibration {
+            calibration.set(load_calibration());
+        }
+    }) as Box<dyn FnMut(web_sys::StorageEvent)>);
+    let _ = window.add_event_listener_with_callback("storage", closure.as_ref().unchecked_ref());
+    closure.forget();
```
CoolCatMona: `closure.forget()` permanently retains this listener. Each page remount adds another callback and captured calibration signal. Please return cleanup that removes the listener.

frontend/src/pages/stones.rs
reid23: positive is correct since a) it should match the other slider and b) right should mean earlier and left should be later like a normal timeline.

going to rename to "audio offset" and add labels to the slider that say this

frontend/Cargo.toml
reid23: added working test recipe to just file and added CI. Not sure if the CI change works but we'll find out when it gets to main and the action actually runs

frontend/src/time_sync.rs
reid23: good catch thanks

Issue #196, opened 2026-05-28
See original match time


reid23:
## Objectives
- make it more clear when a tournament is running behind schedule
- make it more immediately obvious that teams should get to their games asap
- make the idea of the start deadline (currently 'nominal start time') more intuitive

## Key Results
- show the current time on the schedule and timeline (see #195) views
- show matches at their originally scheduled time or earlier, no matter their status. show their end time as at latest their scheduled end time. add a toggle to "show show true times" which moves them to when they really happened and makes them the right length.
- on a match's page, clearly label the planned start. for STATIC and FAST matches, don't show anything else. For SAFE matches, show the start deadline.
- clearly display the status of every match on its page. add documentation through a help button or something to explain what each status means.

## Implementation Plan
will need to add a `scheduled_start` time that records the initial nominal start time. will need to be careful about how editing matches affects this. probably should be what the scheduled order comes from; see #148.
otherwise it's a pretty simple frontend update.

PR #252, opened 2026-08-14
[Feature] toml update, solver fixes, and scheduling docs update

### LINKED ISSUE #243: fix toml import/export
## Objectives
- make it easier to save progress
- make it easier to develop a schedule/bracket design tool that integrates well

## Key Results
- toml import/export of schedule, including all matches, tags, fields, and the bracket diagram
- human readable files: no uuids
- loading from toml clears all existing matches, tags, fields, and bracket diagram elements. 
- loading from toml should not be possible once any match has begun. (for now this is best; we can extend later)

## Implementation Plan

matches: name, field, team1, team2, refs, start_time, prev_match, length, schedule_type, ribbon, nsets, stones_per_set, skip_condition
fields: name
tags: name
brackets: 
match: x, y, w, h, net1p, net2p, flipped, match
text: x, y, size, text
labeledteam: x, y, netp, label, team
image: x, y, w, h, image (base64 probably)

on upload, check basic validity (do all the references exist?), then put it in, and do a solve. see what happens.

make sure to warn "make a backup! this will delete everything!"

reid23:
## Summary

really should be a bugfix title and branch but im kinda lazy...

## What changed

Backend:

- push-back shifts `scheduled_start_time` together with nominal on STATIC anchors (it used to only move nominal, so the published plan silently kept the old times). backfills whichever field is missing so the two stay aligned, and re-solves plan-then-live instead of live-only.
- force-start-to-STATIC writes "now" into *both* start fields and re-solves both passes.
- the form-based TOML import now recomputes both passes after import, matching what the JSON import already did.
- match-detail API exposes `scheduled_start_time` (the frontend needs it later in the stack).

TOML:

- get rid of event key; just use whatever event we're uploading to.
- export includes `scheduled_start_time`; import seeds plan from nominal (and vice versa) for legacy files that only have one.
- unknown teams import as new tags. user still gets warned

Docs:

- `scheduling.md` includes two-pass solve and scheduled_start_time.

## Test plan

- [x] `just lint` + `ruff format --check` clean
- [x] Manual verification:
  - [x] exported a schedule, wiped, re-imported: plan times survive
  - [x] imported a file referencing a fake team: match imports with `tag::` ref + warning

## Linked issues

Refs #196, #243 

---

### Pre-review checklist

- [x] Branch name follows `category/name`.
- [x] Targeting `dev`, not `main`.
- ~~ If a new module was added under `app/`~~
- ~~If developer workflow changes~~

### DIFF
--- a/app/routes/tournaments/scheduling.py
+++ b/app/routes/tournaments/scheduling.py
@@ -185,6 +185,7 @@ def import_schedule(tournament_url):
             "matches_created": import_result.matches_created,
             "matches_updated": import_result.matches_updated,
             "errors": import_result.errors,
+            "warnings": import_result.warnings,
         }
 
     return json_from_result(res, ok_to_payload=result_to_payload)
@@ -1427,9 +1428,9 @@ def import_schedule_api(tournament_url):
     from app.services.schedule_import_export_service import ScheduleImportExportService
     from app.utils.result_helpers import json_from_result
 
-    def _ok_payload(_):
+    def _ok_payload(import_result):
         recompute_scheduled_and_nominal_times(tournament_url)
-        return {}
+        return {"warnings": import_result.warnings}
 
     res = ScheduleImportExportService.import_schedule(tournament_url, toml_content)
     return json_from_result(res, ok_to_payload=_ok_payload)
--- a/app/routes/tournaments/scheduling.py
+++ b/app/routes/tournaments/scheduling.py
@@ -175,7 +175,8 @@ def import_schedule(tournament_url):
     res = ScheduleImportExportService.import_schedule(tournament_url, toml_content)
 
     def result_to_payload(import_result):
-        """Convert ImportResult to JSON payload."""
+        """Convert ImportResult to JSON payload, then re-solve plan + live timelines."""
+        recompute_scheduled_and_nominal_times(tournament_url)
         return {
             "tags_created": import_result.tags_created,
             "tags_updated": import_result.tags_updated,
@@ -1267,9 +1268,13 @@ def force_start_match_api(tournament_url, match_id):
     r_csv, i_csv = resolve_refs_slots(refs_list, tournament_url)
     set_match_referees_from_csv(match, r_csv, i_csv)
 
-    # Convert to static
+    # Convert to static. The new wall-clock "now" becomes both the plan anchor and
+    # the live estimate — otherwise the planned pass keeps a stale scheduled time
+    # while the board (plan mode) still shows the old slot.
     match.schedule_type = ScheduleType.STATIC
-    match.nominal_start_time = now_utc_naive()
+    now = now_utc_naive()
+    match.nominal_start_time = now
+    match.scheduled_start_time = now
     match.status = MatchStatus.READY_TO_START
 
     # The match is being converted to STATIC and started — fully splice it out of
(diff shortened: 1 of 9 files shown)


app/routes/tournaments/scheduling.py
```diff
@@ -1330,12 +1337,25 @@ def push_back_matches_api(tournament_url):
...
     from datetime import timedelta
 
+    # Push-back is a deliberate shift of the *plan* of the day (e.g. late first
+    # whistle). STATIC anchors must move scheduled_start_time as well as nominal;
+    # dynamic matches then re-derive both timelines from the new anchors.
+    delta = timedelta(minutes=minutes)
     for m in matches:
-        if m.schedule_type == ScheduleType.STATIC and m.nominal_start_time:
-            m.nominal_start_time += timedelta(minutes=minutes)
+        if m.schedule_type != ScheduleType.STATIC:
```
CoolCatMona: `READY_TO_START` matches are still unstarted, but the query above excludes them. Push-back therefore leaves ready static anchors unchanged. Please exclude only started matches.

app/services/schedule_import_export_service.py
```diff
@@ -27,28 +28,80 @@
...
+    warnings: list[str] = []
+    rewritten: list[str] = []
+
+    def _repl(m: re.Match) -> str:
+        inner = m.group(1).strip()
+        if not inner or "::" in inner or inner in known_teams:
+            return m.group(0)
+        rewritten.append(inner)
+        warnings.append(f"Team '{inner}' not found; imported as tag reference 'tag::{inner}'")
+        return f"[tag::{inner}]"
```
CoolCatMona, line 121: Unknown teams can become unresolved tags here, but new static matches still initialize as `READY_TO_START`. The solver does not downgrade that status. Please initialize unresolved static matches below READY.

app/routes/tournaments/scheduling.py
```diff
@@ -1408,9 +1428,9 @@ def import_schedule_api(tournament_url):
     from app.services.schedule_import_export_service import ScheduleImportExportService
     from app.utils.result_helpers import json_from_result
 
-    def _ok_payload(_):
+    def _ok_payload(import_result):
         recompute_scheduled_and_nominal_times(tournament_url)
-        return {}
+        return {"warnings": import_result.warnings}
```
CoolCatMona, line 1374: The endpoint now returns warnings, but `api::import_schedule` discards this payload and closes the modal. Organizers never see rewritten-team warnings. Please surface them.

app/routes/tournaments/scheduling.py
```diff
@@ -1330,12 +1337,25 @@ def push_back_matches_api(tournament_url):
...
+            m.nominal_start_time = m.nominal_start_time + delta
+        # Keep the two aligned if only one was populated.
+        if m.scheduled_start_time is None and m.nominal_start_time is not None:
+            m.scheduled_start_time = m.nominal_start_time
+        if m.nominal_start_time is None and m.scheduled_start_time is not None:
+            m.nominal_start_time = m.scheduled_start_time
 
     db.session.commit()
-    recompute_all_match_times(tournament_url)
+    recompute_scheduled_and_nominal_times(tournament_url)
```
CoolCatMona: This route shifts static plan anchors and recomputes. The legacy route shifts nominal and confirmed values for all unstarted matches without recomputing. Please share one implementation.

app/routes/tournaments/scheduling.py
reid23: added

app/routes/tournaments/scheduling.py
reid23: oops you right, yikes. fixed

app/routes/tournaments/scheduling.py
reid23: ok

app/services/schedule_import_export_service.py
reid23: oops. fixed.

Issue #164, opened 2026-05-08
Schedule Display Improvements


reid23:
**Is your feature request related to a problem? Please describe.**
- often you can't see playing and reffing teams on the schedule, so you have to click on the match
- for dynamic scheduling, since matches get pushed back, it's hard to see how far behind schedule you are. 
- it's hard to tell how far away things are in the schedule

**Describe the solution you'd like**
- probably needs #33 first, but there should be an easy way to ensure that team names fit in the schedule block.
- needs some graphic design work but maybe the main schedule blocks shouldn't update to show reality, but rather stay at nominal, and then have separate indicators that show deadlines and true progress
- there should be a line across the schedule indicating the current time, like in google calendar

PR #255, opened 2026-08-14
[Feature] upgrade break match type

reid23:
## Summary

- new static break match type so you can start the day with a break labeled "chain comp" or something without any issues
    - static breaks are marked completed when their start time passes. this is so that they aren't completed immediately on creation because you can't edit completed matches.
- breaks and joins are presented in the editing as a single match. you just select the fields you want.
- breaks act like joins now. if you want separate breaks, you need separate names. 
    - THIS IS NOT BACKWARDS COMPATIBLE. however all past tournaments are done so i think its fine
    - this is done so that we can in the future add team requirements to breaks. i think its better this way. plus now we don't need joins technically. perhaps we can deprecate them in the future or add a note in documentation about this.

note: because of the new schedule views where players just see by team or by field, their primary view (by team) won't show breaks, because breaks only exist in the context of a field. I debated making breaks also require teams (could use the ref table) but:
- it would be confusing to then still require a field to be selected? but we have to because the model requires it and the scheduling solver requires it unless we want a huge redesign there. and the whole edit process is very field-centric so it'd just be confusing to work with.
- the way we indicate priority for use of the "team" resource is by order of scheduled_start_time, which requires a field selection to determine. 
- if we don't have a field selected, we can't see the break anywhere on the schedule editor
- if we have multiple fields and multiple teams in a break, then all the matches across all fields need to have a matching set of entries for required teams. which feels sketchy.  
overall it just felt super scuffed, so im willing to just cope with the fact that the team schedule won't show breaks. at least for now.

really i think it is the correct model to allow any arbitrary selection of teams and/or fields based on what you're actually trying to say: are you giving people time to eat? then just teams. Are you doing bear pits on field 4? then you need all the teams and nothing happening on field 4. etc. but first we need to do a lot of architectural thinking about how to express schedules and how to make a UI that makes that easy to do correctly.

## What changed

Solver:

- same-name breaks across fields act as **one joined break**: every row in the group gets the union of the group's dependency edges, so they all start together (both solver passes), while keeping their own field, length, and chain position. joins already merged like this; now breaks match.
- **STATBREAK**: a break with a fixed start time. the solver never moves it, and its *status* is a pure function of the clock — `Match.effective_status` returns COMPLETED once the start passes, and the stored status is never written. 
- the participant lookup for cross-field conflict edges now reads the `match_referees` table instead of the legacy `refs` CSV column (which dual-write stopped maintaining a while ago — ref-based double-booking detection was quietly broken).

API:

- new structural-group endpoints: `POST/PUT/DELETE /_api/tournaments/<url>/break-groups[/<name>]`. create one row per selected field (chain-tail default, or explicit per-field previous match), edit the whole same-name group at once (length / start time / field membership with chain splicing), delete the group. JOIN groups force length 0.
- breaks and joins occupy *fields only* — team requirements on structural rows are rejected everywhere.

Frontend:

- picking Break/Static Break/Join in the create form switches to the group form: name + field checkboxes (+ length, + start time for STATBREAK), with select all/none. clicking any break or join in the editor opens the group editor for the whole group.

## Migration impact

- Alembic revisions added: `0012_statbreak`. recreates the two partial unique indexes so STATBREAK joins the per-field name-uniqueness group. index-only; downgrade demotes STATBREAK rows to BREAK first.
- Backwards compatible: existing BREAK/JOIN rows untouched.

## Test plan

- [x] `just test` passes
- [x] `just lint` clean, `cargo check` clean
- [x] Manual verification:
  - [x] created a lunch break across 3 fields, dragged the morning around, all three stayed in lockstep
  - [x] STATBREAK at 12:00 sharp: morning running late flows around it, shows completed at 12:00 regardless

## Linked issues

Refs #199 
Refs #164

---

### Pre-review checklist

### DIFF
--- a/app/domain/enums.py
+++ b/app/domain/enums.py
@@ -123,7 +123,11 @@ class ScheduleType(StrEnum):
             delays.
         FAST: Times are recalculated aggressively, scheduling matches as
             early as possible.
-        BREAK: A scheduled break (no match played).
+        BREAK: A scheduled break (no match played). Same-name breaks across
+            fields start together (their dependency edges are unioned).
+        STATBREAK: A statically scheduled break: the user supplies the start
+            time and the solver never moves it. Auto-completes once its
+            window (start + nominal_length) has elapsed.
         JOIN: A synchronisation point that waits for multiple preceding
             matches to complete before advancing.
     """
@@ -132,7 +136,24 @@ class ScheduleType(StrEnum):
     SAFE = "SAFE"
     FAST = "FAST"
     BREAK = "BREAK"
+    STATBREAK = "STATBREAK"
     JOIN = "JOIN"
+
+
+#: Break-like schedule types: no playing teams, may carry ref-slot team
+#: requirements ("these teams must attend the break").
+BREAK_SCHEDULE_TYPES: tuple[ScheduleType, ...] = (
+    ScheduleType.BREAK,
+    ScheduleType.STATBREAK,
+)
+
+#: Structural schedule types: not games (no team1/team2, no lifecycle chrome),
+#: unique per (name, event, field) rather than per (name, event).
+STRUCTURAL_SCHEDULE_TYPES: tuple[ScheduleType, ...] = (
+    ScheduleType.BREAK,
+    ScheduleType.STATBREAK,
+    ScheduleType.JOIN,
+)
 
 
 class SetType(StrEnum):
--- a/app/domain/enums.py
+++ b/app/domain/enums.py
@@ -126,8 +126,9 @@ class ScheduleType(StrEnum):
         BREAK: A scheduled break (no match played). Same-name breaks across
             fields start together (their dependency edges are unioned).
         STATBREAK: A statically scheduled break: the user supplies the start
-            time and the solver never moves it. Auto-completes once its
-            window (start + nominal_length) has elapsed.
+            time and the solver never moves it. Its status is derived from the
+            current time when read (COMPLETED once the start has passed),
+            never stored.
         JOIN: A synchronisation point that waits for multiple preceding
             matches to complete before advancing.
     """
@@ -139,13 +140,6 @@ class ScheduleType(StrEnum):
     STATBREAK = "STATBREAK"
     JOIN = "JOIN"
 
-
-#: Break-like schedule types: no playing teams, may carry ref-slot team
-#: requirements ("these teams must attend the break").
-BREAK_SCHEDULE_TYPES: tuple[ScheduleType, ...] = (
-    ScheduleType.BREAK,
-    ScheduleType.STATBREAK,
-)
 
 #: Structural schedule types: not games (no team1/team2, no lifecycle chrome),
 #: unique per (name, event, field) rather than per (name, event).
--- a/app/models/match.py
+++ b/app/models/match.py
@@ -91,14 +91,14 @@ class Match(db.Model):
             "event",
             "field",
             unique=True,
-            sqlite_where=sa.text("schedule_type IN ('BREAK', 'JOIN')"),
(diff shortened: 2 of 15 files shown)


app/routes/tournaments/matches_admin.py
```diff
@@ -683,6 +747,357 @@ def delete_match_api(tournament_url, match_id):
...
+        return None
+    if dt.tzinfo:
+        dt = dt.astimezone(timezone.utc).replace(tzinfo=None)
+    return dt
+
+
+def _break_group_rows(tournament_url: str, name: str) -> list[Match]:
+    """All BREAK/STATBREAK/JOIN rows sharing *name* in this tournament."""
+    return (
+        Match.query.filter_by(event=tournament_url, name=name)
```
CoolCatMona, line 771: This groups structural rows by display name alone. Existing unrelated rows with the same name will merge, and mixed types make `rows[0]` unreliable. Please use a stable group ID.

app/models/match.py
```diff
@@ -219,6 +221,22 @@ def loser_team_id(self) -> str | None:
...
+        """Lifecycle status as clients should see it.
+
+        For ``STATBREAK`` the status is a pure function of the current time —
+        ``COMPLETED`` once the scheduled start has passed, ``NOT_STARTED``
+        before that — and the stored :attr:`status` is ignored (the solver
+        never writes it). All other schedule types return the stored status.
+        """
+        if self.schedule_type == ScheduleType.STATBREAK:
+            start = self.nominal_start_time or self.scheduled_start_time
+            if start is not None and now_utc_naive() >= start:
```
CoolCatMona, line 238: A STATBREAK becomes complete at its start, not at `start + nominal_length`. Dependents can become ready while the break is active. Please complete it at the scheduled end.

app/routes/tournaments/matches_admin.py
```diff
@@ -635,15 +698,16 @@ def create_match_api(tournament_url):
             ScheduleType.FAST,
             ScheduleType.STATIC,
             ScheduleType.BREAK,
+            ScheduleType.STATBREAK,
```
CoolCatMona, line 684: STATBREAK reaches this link update without the same-field check used above. A caller can therefore create a cross-field chain. Please validate the predecessor field.

app/routes/tournaments/matches_admin.py
```diff
@@ -683,6 +747,357 @@ def delete_match_api(tournament_url, match_id):
...
+        created.append(match)
+
+    db.session.flush()
+    db.session.commit()
+    recompute_scheduled_and_nominal_times(tournament_url)
+
+    return jsonify({"success": True, "name": name, "uuids": [m.uuid for m in created]})
+
+
+@bp.route("/tournaments/<tournament_url>/break-groups/<name>", methods=["PUT"])
```
CoolCatMona, line 962: `match_name_char_error` permits `/`, but this route matches one path segment. A group named `Lunch/A` can be created but cannot be updated or deleted. Please use a stable ID or reject `/`.

app/domain/enums.py
```diff
@@ -132,9 +137,19 @@ class ScheduleType(StrEnum):
     SAFE = "SAFE"
     FAST = "FAST"
     BREAK = "BREAK"
+    STATBREAK = "STATBREAK"
```
CoolCatMona, line 140: Please document the STATBREAK lifecycle and same-name grouping rules in `docs/scheduling.md`. `schedule-display-vision.md` still lists STATBREAK as a non-goal.

app/models/match.py
reid23: no, we want this behavior. when a break starts, we should be allowed to start the next match early (assuming all other dependencies are satisfied).

PR #256, opened 2026-08-14
[Feature] schedule editing overhaul

### LINKED ISSUE #199: Fix Schedule Editing

### LINKED ISSUE #246: Dependency Display
## Objectives
- make it easier to understand why schedules have solved the way they have
- make it easier to design a schedule in a reasonable way

## Key Results
- functionality like the "ratsnest" in the bracket builder: illustrate with color-coded lines all the dependencies of a given match
- highlight the matches this match depends on with hover + some button

### LINKED ISSUE #247: Reasonable Schedule Entry UI
## Objectives
- entering schedules is annoying and difficult

## Key Results
- drag and drop matches into place on the schedule
- click + drag to create match at time (like gcal)
- see schedule while editing
- more info present in edit mode (match type, etc)
- maybe: separate edit mode from display?

## Implementation Plan
ugh idk

reid23:
## Summary

schedule editing moves off the public page onto its own full-width editor at `/:url/schedule/edit` (linked from the admin card), where you build the schedule by direct manipulation instead of forms-first. generally should be far more intuitive and easy to make changes, especially when it comes to reordering matches.

also fixes two old solver-adjacent bugs that editing kept running into

## What changed

The editor page:

- full viewport width; the match/break/join forms are a **docked card** (right of the schedule on wide screens, above on narrow) instead of modals, so the schedule stays visible and interactive while you edit.
- **drag empty space to create** (gcal style): placeholder follows the drag, releases into the create card prefilled with field/start/length. the placeholder stays on the timeline while the card is open (all checked fields, for break/join groups) and tracks your edits. the "+ Match" button is gone — the schedule *is* the input.
- **drag blocks to move**: STATIC places freely, STATBREAK groups shift their start, dynamic matches snap into the gap after their new previous match with the insertion point highlighted. bad drops error and snap back.
- "Default new match type" dropdown seeds the create card (all six types; structural types open the group form).
- **bulk change length**: click-select blocks, apply one length (new endpoint; locked/join rows reported as skipped).
- **push back day** tool: shift the plan of everything not-yet-started by N minutes (negative pulls forward). the endpoint existed with zero UI; now it has one. the manual "Recompute Times" button and its endpoints are *removed* — the solver re-derives everything on every edit now, so the button had nothing left to fix.
- locked (started/finished) matches of every type stay clickable/draggable/selectable and just fail with a clear message, instead of mysteriously ignoring you. breaks show their real statuses here (public views keep the neutral look).
- with a team input focused, hovering a match shows **Winner/Loser chips** that insert `Name::winner`/`::loser` into the field.
- Alt-hover dependency view: hold alt while hovering a match to see lines to the matches it depends on. only includes previous match, team1, team2, and refs when they're specified as references; no skip condition and no "other match with higher priority is using this team" display. perhaps that can be added later. especially in pod schedules that's nice, but it could be confusing since if you move a match before one of those dependencies it's fine and the matches just reorder. (then again same with prev/next). idk. i digress. that's for later.

The public page:

- identical for TOs and players now: by-team and by-field only, no warnings button, no edit anything. TOs get to the editor through the admin card.

Bugfixes (the editing work surfaced both):

- **reorder chain splice**: moving A in front of B could leave both at the same time. half-linked chains (successor points back, predecessor's `next_match` unset — the normal state around STATIC matches and TOML imports) made the splice helper miss the old successor, so both matches shared a dependency. successor is now derived from back-pointers as a fallback.
- **stale solver statuses**: the live PROCEDURE only ever *upgraded* statuses, so with A(done)→B→C, moving C in front of B left B READY_TO_START forever. solver-earned statuses (READY_TO_START / TIME_FINALIZED / structural COMPLETED) are now re-earned on every solve and downgraded to the type's floor when their justification disappears. real-world statuses (in progress / finalized / skipped) are never touched. topological order makes downgrades cascade in one pass.

## Test plan

- [x] `just test` passes
- [x] `cargo check` clean; dep-line geometry unit tests pass natively
- [x] Manual verification:
  - [x] built a day from scratch by dragging only
  - [x] reordered mid-tournament (some matches done): statuses and times follow correctly
  - [x] alt-hover on a bracket-heavy schedule: rings/lines appear and *disappear* (yes, really — the sticky-outline bug was an inline box-shadow that survived style diffing; rings are a mounted/unmounted child element now)

## Linked issues

Closes #199, #246, #247

---

### Pre-review checklist

### DIFF
--- a/app/routes/tournaments/__init__.py
+++ b/app/routes/tournaments/__init__.py
@@ -48,6 +48,34 @@ def _lookup_match_in(uid: str | None, tournament_url: str) -> Match | None:
     return Match.query.filter_by(uuid=uid, event=tournament_url).first()
 
 
+def _chain_successor(match: Match, tournament_url: str, exclude_uuid: str | None = None) -> Match | None:
+    """The match that sits after *match* in its per-field chain.
+
+    Trusts ``match.next_match`` when set. Otherwise falls back to the unique
+    match whose ``previous_match`` back-pointer aims at *match* — chains can be
+    "half-linked" (child points at parent, parent's ``next_match`` unset), e.g.
+    around STATIC matches (which are deliberately detached on edit) or after a
+    TOML import that only populates ``previous_match``. If several candidates
+    point back (already-degenerate state), returns ``None`` rather than guess.
+
+    Args:
+        match: Chain node whose successor we want.
+        tournament_url: Tournament scope.
+        exclude_uuid: Optional uuid to ignore among back-pointer candidates
+            (used to exclude the match currently being spliced).
+    """
+    if match.next_match:
+        nxt = _lookup_match_in(match.next_match, tournament_url)
+        if nxt is not None:
+            return nxt
+    candidates = [
+        m
+        for m in Match.query.filter_by(previous_match=match.uuid, event=tournament_url).all()
+        if m.uuid != exclude_uuid
+    ]
+    return candidates[0] if len(candidates) == 1 else None
+
+
 def detach_match_from_chain(match: Match, tournament_url: str) -> None:
     """Remove *match* from its per-field doubly-linked chain.
 
@@ -95,18 +123,19 @@ def update_match_previous_link(match: Match, prev_match_id: str, tournament_url:
     if not is_new:
         detach_match_from_chain(match, tournament_url)
 
-    # Splice *match* between prev_match and whatever currently sits after prev_match.
-    new_next_id = prev_match.next_match
-    if new_next_id == match.uuid:
-        # Stale back-pointer guard.
-        new_next_id = None
+    # Splice *match* between prev_match and whatever currently sits after
+    # prev_match. The successor is derived via _chain_successor rather than
+    # prev_match.next_match alone: half-linked chains (successor's
+    # previous_match set, prev's next_match unset — STATIC matches, TOML
+    # imports) would otherwise leave the old successor also pointing at
+    # prev_match, giving two matches the same dependency and thus the same
+    # solved start time.
+    new_next = _chain_successor(prev_match, tournament_url, exclude_uuid=match.uuid)
     match.previous_match = prev_match.uuid
-    match.next_match = new_next_id
+    match.next_match = new_next.uuid if new_next is not None else None
     prev_match.next_match = match.uuid
-    if new_next_id:
-        new_next = _lookup_match_in(new_next_id, tournament_url)
-        if new_next is not None:
-            new_next.previous_match = match.uuid
+    if new_next is not None:
+        new_next.previous_match = match.uuid
 
 
 def delete_matches_with_children(match_uuids: list[str]) -> None:
--- a/app/routes/tournaments/matches_admin.py
+++ b/app/routes/tournaments/matches_admin.py
@@ -747,6 +747,70 @@ def delete_match_api(tournament_url, match_id):
     return jsonify({"success": True})
 
 
+@bp.route("/tournaments/<tournament_url>/matches/bulk-length", methods=["POST"])
+@login_required
+def bulk_match_length_api(tournament_url):
+    """Set ``nominal_length`` on many matches at once.
+
+    Body: ``{"match_ids": [...], "length": <minutes>}``. Applies the shared
+    length to every editable match, skipping JOINs (structurally zero-length)
+    and matches locked by having started (same rule as single-match edit:
+    STATBREAKs stay editable even when COMPLETED). Single commit, one
+    recompute, per-match results in the response.
+    """
+    if not _check_to(tournament_url):
+        return jsonify({"error": "Forbidden"}), 403
+
+    data = request.get_json()
+    if not data:
+        return jsonify({"error": "Invalid JSON"}), 400
+
+    match_ids = data.get("match_ids")
+    if not isinstance(match_ids, list) or not match_ids:
+        return jsonify({"error": "match_ids must be a non-empty list."}), 400
+
+    try:
+        length = int(data.get("length"))
+    except (TypeError, ValueError):
+        return jsonify({"error": "Length must be a number of minutes."}), 400
+    if length <= 0:
+        return jsonify({"error": "Length must be greater than zero."}), 400
+
+    # Dedup while preserving order so results line up with the request.
+    seen: set[str] = set()
+    unique_ids: list[str] = []
+    for mid in match_ids:
+        mid = str(mid or "").strip()
+        if mid and mid not in seen:
+            seen.add(mid)
+            unique_ids.append(mid)
+    if not unique_ids:
+        return jsonify({"error": "match_ids must be a non-empty list."}), 400
+
+    results = []
+    updated = 0
+    for mid in unique_ids:
+        match = Match.query.filter_by(uuid=mid, event=tournament_url).first()
+        if not match:
+            results.append({"match_id": mid, "status": "not_found"})
+            continue
+        if match.schedule_type == ScheduleType.JOIN:
+            results.append({"match_id": mid, "status": "skipped_join"})
+            continue
+        if match.status in _LOCKED_STATUSES and match.schedule_type != ScheduleType.STATBREAK:
+            results.append({"match_id": mid, "status": "skipped_locked"})
+            continue
+        match.nominal_length = length
+        updated += 1
+        results.append({"match_id": mid, "status": "updated"})
+
+    db.session.commit()
+    if updated:
+        recompute_scheduled_and_nominal_times(tournament_url)
+
+    return jsonify({"success": True, "updated": updated, "results": results})
+
+
 # ---------------------------------------------------------------------------
 # Structural groups ("break-groups" API): same-name BREAK/STATBREAK/JOIN rows
 # across multiple fields, created and edited as one unit (shared name / length /
--- a/app/utils/scheduling.py
+++ b/app/utils/scheduling.py
@@ -232,12 +232,50 @@ def _procedure_with_match(
         # BREAK's for time purposes.
         return
 
-    if node.status in (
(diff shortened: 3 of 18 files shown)


app/routes/tournaments/__init__.py
```diff
@@ -48,6 +48,34 @@ def _lookup_match_in(uid: str | None, tournament_url: str) -> Match | None:
     return Match.query.filter_by(uuid=uid, event=tournament_url).first()
 
 
+def _chain_successor(match: Match, tournament_url: str, exclude_uuid: str | None = None) -> Match | None:
+    """The match that sits after *match* in its per-field chain.
+
+    Trusts ``match.next_match`` when set. Otherwise falls back to the unique
+    match whose ``previous_match`` back-pointer aims at *match* — chains can be
```
CoolCatMona, line 51: This helper handles half-linked chains during insertion, but `detach_match_from_chain` still reads only `match.next_match`. Moving a half-linked middle node leaves its child pointing to the moved match. Please use `_chain_successor` during detach.

frontend/src/pages/schedule.rs
```diff
@@ -1078,24 +1216,29 @@ pub fn Schedule(url: String, view: String, team: String, field: String) -> Eleme
...
-                                                    let mut trigger = refresh_trigger;
-                                                    spawn(async move {
-                                                        if let Ok(_) = api::recompute_schedule(&u).await {
-                                                            trigger.set(trigger() + 1);
-                                                        }
-                                                    });
-                                                },
-                                                "Recompute Times"
+                                                class: if push_back_open() { "btn btn-sm btn-primary" } else { "btn btn-sm btn-outline-primary" },
+                                                title: "Shift every not-yet-started match's planned time by N minutes (e.g. the day started late)",
```
CoolCatMona: The UI promises every unstarted match, but the API selects only `NOT_STARTED` and `TIME_FINALIZED`. It then shifts only STATIC rows. `READY_TO_START` anchors and future STATBREAK anchors stay put. Please include both.

app/routes/tournaments/matches_admin.py
```diff
@@ -747,6 +747,70 @@ def delete_match_api(tournament_url, match_id):
...
+        mid = str(mid or "").strip()
+        if mid and mid not in seen:
+            seen.add(mid)
+            unique_ids.append(mid)
+    if not unique_ids:
+        return jsonify({"error": "match_ids must be a non-empty list."}), 400
+
+    results = []
+    updated = 0
+    for mid in unique_ids:
```
CoolCatMona: Bulk updates operate on individual UUIDs. Selecting one row from a multi-field BREAK or STATBREAK changes only that row and splits the shared duration. Please expand structural selections to the complete group or reject them.

app/routes/tournaments/matches_admin.py
```diff
@@ -747,6 +747,70 @@ def delete_match_api(tournament_url, match_id):
...
+        match = Match.query.filter_by(uuid=mid, event=tournament_url).first()
+        if not match:
+            results.append({"match_id": mid, "status": "not_found"})
+            continue
+        if match.schedule_type == ScheduleType.JOIN:
+            results.append({"match_id": mid, "status": "skipped_join"})
+            continue
+        if match.status in _LOCKED_STATUSES and match.schedule_type != ScheduleType.STATBREAK:
```
CoolCatMona: This explicitly bypasses the STATBREAK history lock. A completed static break can be rewritten here although the group endpoint rejects that edit. Please apply the same `effective_status` check.

frontend/src/pages/schedule.rs
```diff
@@ -1379,29 +1781,148 @@ fn CreateMatchModal(
...
+        move || t
+    });
+    let length = use_signal(move || prefill_length.unwrap_or(60));
+    let start_time = use_signal({
+        let s = prefill_start_time.clone();
+        move || s.unwrap_or_default()
+    });
+    // Dynamic default types start with the drag-derived previous match preselected.
+    let mut previous_match_id = use_signal({
+        let init = if matches!(initial_type.as_str(), "SAFE" | "FAST") {
```
CoolCatMona: The drag computes a predecessor, but this initialization discards it for BREAK and JOIN. The group request also omits the predecessor map. Structural drops therefore append to field tails instead of the placeholder position. Please pass the per-field predecessor.

app/utils/scheduling.py
```diff
@@ -232,13 +232,51 @@ def _procedure_with_match(
...
+    # their COMPLETED is solver-derived (nothing else can set it on structural
+    # rows), so it must be re-earned each pass like any other solver status.
+    if node.schedule_type not in (ScheduleType.BREAK, ScheduleType.JOIN) and node.status in (
         MatchStatus.COMPLETED,
         MatchStatus.IN_PROGRESS,
         MatchStatus.SKIPPED,
     ):
         return
 
+    # --- Downgrade pass -----------------------------------------------------
```
CoolCatMona, line 248: This downgrade runs only for topologically ordered nodes. Cycle nodes use `_procedure_for_cycle_node`, which preserves solver-derived statuses. A new cycle can therefore leave a match `READY_TO_START`. Please reset these statuses in the cycle path.

frontend/src/pages/schedule.rs
```diff
@@ -1187,6 +1477,84 @@ pub fn Schedule(url: String, view: String, team: String, field: String) -> Eleme
...
+                                                };
+                                                api::update_match(&u, &mc.match_id, &req).await
+                                            }
+                                            "STATBREAK" => {
+                                                // Static breaks move as a group: shared start time.
+                                                let req = UpdateBreakGroupRequest {
+                                                    schedule_type: None,
+                                                    length: None,
+                                                    start_time: mc.new_start_utc.clone(),
+                                                    fields: None,
```
CoolCatMona, line 1622: The drag payload includes the destination field, but this request sends `fields: None`. Horizontal STATBREAK moves save only the time and snap back after refresh. Please send the updated fields or disable horizontal dragging.

docs/scheduling.md
```diff
@@ -39,7 +39,7 @@ mode has an option to display times exactly as they happened.
 |-------|------|
 | Match start / end / finalize | Live only → `recompute_all_match_times` (= `run_scheduling(scheduled_pass=False)`) |
 | Match create / edit / delete | Both → `recompute_scheduled_and_nominal_times` |
-| TO “Recompute times” / TOML import (API) / app boot | Both |
+| Push-back / TOML import / app boot | Both |
```
CoolCatMona, line 42: Please document the new editing, grouping, and STATBREAK behavior here. `schedule-display-vision.md` still lists schedule editing and STATBREAK as non-goals.

app/routes/tournaments/matches_admin.py
reid23: got rid of uuids but fixed. uses the whole group

frontend/src/pages/schedule.rs
reid23: added

app/utils/scheduling.py
reid23: cycles are fake and shouldnt really matter but fair enough for consistency's sake. added

frontend/src/pages/schedule.rs
reid23: fixed

frontend/src/pages/schedule.rs
reid23: oops, included

PR #251, opened 2026-08-12
[Feature] ASS usability improvements

reid23:
## Summary

make ASS usable, feature-wise.

(let me have my fun :bangbang: i know its useless!)

## What changed

entry box component changes:
- corresponding paren highlighting
- make autocomplete usable
- reasonable navigation of multi-line text

big ass changes:
- `let`
- `cond` (chained/multi-branch conditionals)
- `won?`
- `wins` and `losses` over a list of matches
- `points-won` and `points-lost` over a list of matches
- `list` dynamic list constructor
- `cons` (prepend)
- `append` (really concat)
- `empty?` check if list empty
- `member?` check if elem in list
- `range` like python
- `filter` (filter list by predicate)
- `map-indexed`
- `sort-by` (one or more key functions)

small ass changes:
- `min_by` -> `min-by`
- `max_by` -> `max-by`

## Migration impact

none because nobody has ever used `min_by` and `max_by` so we're chill

## Test plan

- [x] `just test` passes locally
- [x] `just lint` passes locally, or CI coverage is sufficient
- [x] `just coverage-check` passes locally, or CI coverage is sufficient
- [x] Manual verification:
  - wrote a bunch of fun scripts and they all seemed fine

---

### Pre-review checklist

- [x] Title prefixed with one of `[Feature]`, `[Bugfix]`, `[Refactor]`, `[Documentation]`.
- [x] Branch name follows `category/name` (`feat/...`, `bugfix/...`, `refactor/...`).
- [x] Targeting `dev`, not `main` (unless this is a release PR).
- [x] `just lint` and `just format` are clean.
- [x] Tests added or updated for the new behavior.
- ~~If a new module was added under `app/`: corresponding `app/<area>/README.md` and `docs/api/*.rst` are updated.~~
- ~~If developer workflow changes: `README.md`, `CONTRIBUTING.md`, or `TESTING.md` updated.~~
- [x] No merge conflicts with the base branch.

Issue #215, opened 2026-06-08
More User-Friendly ASS UI


reid23:
**Is your feature request related to a problem? Please describe.**
see the discussion [here](https://github.com/reid23/arctos/pull/205#issuecomment-4635532193)

tldr making the user write ASS is lwk not very good. its hard. its intimidating.

**Describe the solution you'd like**
skip conditions and other future things that use ASS should have a more user-friendly UI. To preserve the complete expressiveness of ASS we can have a "expert mode" or something that allows the raw ASS editing with the existing UI, but for the most part that should be abstracted away.

**Describe alternatives you've considered**
see the discussion liked above

Issue #13, opened 2025-12-22
proper foreign key constraint for fields


reid23:
right now the field attribute of a match is just a string, which could theoretically cause issues, especially once toml import is implemented.

- migrate to requiring a foreign key constraint
- adjust deletion & editing of fields so that nothing bad happens
- write code to migrate all current matches from what we're currently doing to the new version
- do actual db migration (manual) (requires scheduled downtime for the site)

CoolCatMona:
Assuming matches.field should be updated to matches.field_id, which is the foreign key to the `Field`, which requires a migration/backfill.
Can do the same with `cameras`
- find matches whose field doesn't match any Field row in their tournament.
- Backfill via subquery `BATCH_ALTER_TABLE`

Editing of the fields shouldn't matter(?)
Deleting a field referenced by a camera/match should surface an error

References to `match.field` and `camera.field` would just need to be updated to whatever the new fk is

PR #257, opened 2026-08-14
[Feature] ass tag definitions & global variables.

reid23:
## Summary

- tags can now be *defined* by an ASS expression instead of hand-assigned (can still be manually overridden)
- tournaments get named script variables usable in any expression. 

the idea is that:
- lots of logic (ie, seeding logic) can be done automagically now, combined with #251  
- this provides a way to set teams with ASS expressions without breaking the solver's dependency analysis system and without doing really complicated formal analysis of the ASS to figure out what the true dependencies are

## What changed

Tags:

- a tag may carry an ASS expression whose result type must be TEAM (e.g. `(winner {Semi A})`). the existing team dropdown becomes an **override** that always wins when set. resolution order: manual override → evaluate expression → unresolved.
- resolution is centralized in `resolve_tag_to_team` with a cycle guard (tag expressions referencing `[tag::Other]` in a loop resolve to nothing instead of recursing forever), so slot resolution, solver readiness, and `[tag::X]` literals inside other expressions all pick it up for free. tags resolve themselves as results come in — no new triggers needed.
- write-time validation: expression must parse, pass static checks, and infer TEAM. the tags modal gets a validated expression editor with an effective-resolution hint.

Script variables:

- named variables, one expression each, usable as identifiers in *any* ASS expression in the tournament — skip conditions, tag definitions, other variables. lambda-valued variables are callable.
- evaluated eagerly in topological order into the interpreter env. cycles are rejected at write time (and degrade safely at eval time); names must be valid identifiers and can't shadow builtins.
- managed from a new Scripting modal on the schedule editor.

TOML:

- variables round-trip through export/import (`[[variables]]` section, written before `[[tags]]` since tag/match expressions may reference them; reconciled like tags: update by name, create, delete missing; validated with the same rules as the CRUD endpoints).
- found and fixed while verifying: tag expressions were *imported* but never *exported* — they silently vanished from exported files. also fixed import validation errors crashing with an AttributeError instead of reporting (`Err` stores `.val`, not `.value` :skull:).
- PR 1's unknown-team rewriter extends to the new surfaces: `[Team]` literals inside imported tag expressions and variable expressions get the same `tag::` rewrite + warning.

Docs:

- short "Tags as Expressions" and "Variables" sections in the ASS docs.

## Migration impact

- Alembic revisions added: #0013 (`0013_tag_expr_script_vars`). adds nullable `tags.expression` and the `script_variables` table.
- Backwards compatible: tags without expressions behave exactly as before.

## Test plan

- [x] `just test` passes (521; 39 new: resolution order, cycles, `_slot_resolved` integration, variable topo/eval/cycle handling, a skip-condition-via-variable actually skipping through the solver, TOML round-trips + rewriter + idempotence)
- [x] `just lint` clean, `cargo check` clean
- [x] Manual verification:
  - [x] defined `Seed 1` as `(winner {Semi A})`, finalized Semi A, watched the dependent match go ready
  - [x] override beats expression; clearing the override falls back
  - [x] export → wipe → import: expressions and variables intact

## Linked issues

Refs #215, #243

---

### Pre-review checklist

### DIFF
--- a/app/routes/tournaments/matches_admin.py
+++ b/app/routes/tournaments/matches_admin.py
@@ -44,6 +44,7 @@ from models import (
     Field,
     Match,
     Point,
+    ScriptVariable,
     Tag,
     db,
 )
@@ -1206,6 +1207,33 @@ def delete_field_api(tournament_url, field_id):
     return jsonify({"success": True})
 
 
+def validate_tag_expression(tournament_url, expression):
+    """Validate an ASS tag expression: parse + type check (must include TEAM).
+
+    Returns ``(normalized_expression_or_None, error_or_None)``. An empty or
+    None expression normalizes to None (clears the tag's expression).
+    """
+    from app.utils.parser import DSLValidationError, _human_type_name, _infer_types, get_parser
+
+    expression = (expression or "").strip()
+    if not expression:
+        return None, None
+    try:
+        parser = get_parser(tournament_url)
+        warnings = parser.static_check(expression)
+        if warnings:
+            return None, "; ".join(warnings)
+        result = parser.parse(expression)
+    except DSLValidationError as e:
+        return None, str(e)
+    except Exception as e:
+        return None, f"Parse error: {e}"
+    types = _infer_types(result)
+    if "TEAM" not in types:
+        return None, f"Tag expression must resolve to a TEAM, got {_human_type_name(types)}."
+    return expression, None
+
+
 @bp.route("/tournaments/<tournament_url>/tags", methods=["POST"])
 @login_required
 def create_tag_api(tournament_url):
@@ -1225,7 +1253,11 @@ def create_tag_api(tournament_url):
     if Tag.query.filter_by(event=tournament_url, name=name).first():
         return jsonify({"error": "Tag already exists"}), 400
 
-    tag = Tag(event=tournament_url, name=name)
+    expression, err = validate_tag_expression(tournament_url, data.get("expression"))
+    if err:
+        return jsonify({"error": err}), 400
+
+    tag = Tag(event=tournament_url, name=name, expression=expression)
     db.session.add(tag)
     db.session.commit()
     return jsonify({"success": True, "id": tag.id})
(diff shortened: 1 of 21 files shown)


reid23:
@CoolCatMona i requested review on this just because if you could check out this branch and mess around with the schedule editing and viewing UI and lmk if theres any big gaps that'd be super super awesome.

ik its a lot of code in this stack; ive tested it a bunch and im pretty sure its good. just need UI

CoolCatMona:
Doing this today

app/utils/scheduling.py
```diff
@@ -160,7 +160,15 @@ def _slot_resolved(
...
             tag = Tag.query.filter_by(event=tournament_url, name=tag_name).first()
-        return bool(tag and getattr(tag, "team", None))
+        if tag is None:
+            return False
+        if getattr(tag, "team", None):
+            return True
+        if (getattr(tag, "expression", None) or "").strip():
+            from app.utils.helpers import resolve_tag_to_team
+
+            return resolve_tag_to_team(f"tag::{tag_name}", tournament_url) is not None
```
CoolCatMona, line 170: An expression-backed tag can resolve here after its source match finishes, but this path does not write the team into match or referee columns. The match can become `READY_TO_START` while start eligibility still sees empty participants. Please reconcile tag-backed slots before promotion.

app/utils/parser.py
```diff
@@ -1835,9 +1984,11 @@ def parse(text):
             except LarkError as e:
                 raise DSLValidationError(f"Parse error: {e}") from e
             interpreter = Simplifier(
-                lambda x: parse_team_literal(x, event),
+                parse_team,
                 parse_match,
-                env=dict(base_env),
+                # Script variables first, explicit caller env on top (callers
+                # binding globals shouldn't be shadowed by user variables).
+                env={**variable_env(), **base_env},
```
CoolCatMona, line 2101: Script variables are expanded only during evaluation. The dependency analyzer still receives the original skip expression, so a variable containing `(is-skipped {Semi A})` creates no graph edge to `Semi A`. That consumer can become ready too early. Please expand variables during dependency analysis.

app/routes/tournaments/matches_admin.py
```diff
@@ -1268,6 +1300,142 @@ def update_tag_api(tournament_url, tag_id):
...
+    data = request.get_json() or {}
+    name = (data.get("name") or var.name).strip()
+    expression = (data.get("expression") or var.expression).strip()
+    err = _validate_script_variable(tournament_url, name, expression, exclude_id=var.id)
+    if err:
+        return jsonify({"error": err}), 400
+
+    var.name = name
+    var.expression = expression
+    db.session.commit()
```
CoolCatMona, line 1554: These variable mutations commit without participant reconciliation or schedule recomputation. A ready match can keep stale status after its variable changes. Please recompute derived state after create, update, and delete.

app/routes/tournaments/matches_admin.py
```diff
@@ -1268,6 +1300,142 @@ def update_tag_api(tournament_url, tag_id):
...
+
+    var = ScriptVariable.query.filter_by(id=var_id, event=tournament_url).first_or_404()
+    data = request.get_json() or {}
+    name = (data.get("name") or var.name).strip()
+    expression = (data.get("expression") or var.expression).strip()
+    err = _validate_script_variable(tournament_url, name, expression, exclude_id=var.id)
+    if err:
+        return jsonify({"error": err}), 400
+
+    var.name = name
```
CoolCatMona, line 1552: Renaming or deleting a variable does not check consumers in other variables, tag expressions, or skip conditions. Those expressions can silently retain an unknown identifier. Please block the operation or update consumers atomically.

app/routes/tournaments/matches_admin.py
```diff
@@ -1268,6 +1300,142 @@ def update_tag_api(tournament_url, tag_id):
     data = request.get_json()
     if not data or "name" not in data:
         return jsonify({"error": "Name required"}), 400
+    if "expression" in data:
+        expression, err = validate_tag_expression(tournament_url, data.get("expression"))
+        if err:
+            return jsonify({"error": err}), 400
+        tag.expression = expression
```
CoolCatMona, line 1417: Now that tags can appear inside tag and variable expressions, `_tag_usage` does not cover every consumer. Rename or delete can orphan those expressions. Please use an AST-based consumer check.

migrations/versions/0014_tag_expressions_script_variables.py
```diff
@@ -0,0 +1,53 @@
...
+
+def upgrade() -> None:
+    op.add_column("tags", sa.Column("expression", sa.Text(), nullable=True))
+    op.create_table(
+        "script_variables",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("event", sa.String(length=100), nullable=False),
+        sa.Column("name", sa.String(length=100), nullable=False),
+        sa.Column("expression", sa.Text(), nullable=False),
+        sa.ForeignKeyConstraint(["event"], ["tournaments.url"]),
```
CoolCatMona, line 44: This foreign key has no delete cascade, and permanent tournament deletion does not remove script variables first. With SQLite foreign keys enabled, deleting a tournament with variables can fail. Please add cascade behavior or explicit cleanup.

app/routes/tournaments/read.py
```diff
@@ -780,9 +781,27 @@ def tournament_schedule_setup(tournament_url):
...
+            "team": t.team,
+            "expression": t.expression,
+            "resolved_team": resolve_tag_to_team(f"tag::{t.name}", tournament_url),
+        }
+        for t in tags_query
+    ]
+
+    # Tournament-scoped ASS script variables (for the Scripting modal).
+    script_variables_data = [
+        {"id": v.id, "name": v.name, "expression": v.expression}
```
CoolCatMona: Published schedules can call this endpoint without authentication, but this response includes raw tag expressions and every script variable. The dedicated variable route is organizer-only. Please return raw scripting data only to organizers.

app/routes/tournaments/read.py
```diff
@@ -780,9 +781,27 @@ def tournament_schedule_setup(tournament_url):
...
+
     tags_query = Tag.query.filter_by(event=tournament_url).order_by(Tag.name).all()
-    tags_data = [{"id": t.id, "name": t.name, "team": t.team} for t in tags_query]
+    tags_data = [
+        {
+            "id": t.id,
+            "name": t.name,
+            "team": t.team,
+            "expression": t.expression,
+            "resolved_team": resolve_tag_to_team(f"tag::{t.name}", tournament_url),
```
CoolCatMona, line 781: This resolves every tag separately. Each call creates a parser and evaluates all script variables, while the outer cache clears after each resolution. Public schedule loads can repeat substantial parser and database work. Please reuse one request-scoped evaluation context.

app/utils/parser.py
```diff
@@ -1700,6 +1706,126 @@ def _evaluate_min_by(self, head, args):
...
+        return names
+    for child in tree.children:
+        names |= _extract_identifier_names(child)
+    return names
+
+
+def extract_variable_references(text: str) -> set[str]:
+    """Names of identifiers in `text` that could reference script variables.
+
+    Builtins and keyword literals are excluded. Lambda parameter names are NOT
```
CoolCatMona: This extractor treats lambda parameters, `let` bindings, and quoted identifiers as global dependencies. Valid lexical expressions can then fail the cycle check. Please collect only free identifiers.

app/utils/toml_helpers.py
```diff
@@ -48,6 +51,11 @@ def parse_toml_schedule(content: str) -> Result[dict[str, Any], ArctosError]:
     if not isinstance(event, str) or not event:
         event = None
 
+    # Extract script variables (optional, defaults to empty list)
+    variables = data.get("variables", [])
+    if not isinstance(variables, list):
```
CoolCatMona, line 56: `variables = [1]` passes this list check, then the rewrite path calls `.get()` on the integer before serializer validation. That produces a server error. Please require every `variables` entry to be a table here.

app/utils/helpers.py
```diff
@@ -236,9 +291,14 @@ def resolve_tag_to_team(tag_ref: str, tournament_url: str) -> str | None:
...
-    if tag and tag.team:
+    if not tag:
+        return None
+    if tag.team:
         return tag.team
-    return None
+    expression = (getattr(tag, "expression", None) or "").strip()
+    if not expression:
+        return None
+    return _resolve_tag_expression_to_team(tag.name, expression, tournament_url)
```
CoolCatMona, line 301: This helper is now the effective tag resolver, but bracket payloads still read `Tag.team` directly. Expression-backed tags can resolve on the schedule and remain unresolved on the bracket. Please use this resolver in every bracket path.

app/utils/parser.py
```diff
@@ -1700,6 +1706,126 @@ def _evaluate_min_by(self, head, args):
...
+# --------------------------------------------------------------------------
+
+# Identifier names that can never be used as script-variable names: builtin
+# functions, special forms, and the keyword literals.
+RESERVED_IDENTIFIERS = frozenset(Simplifier.BUILTINS) | {"true", "false", "nil"}
+
+# Mirrors the IDENTIFIER terminal in grammar.lark.
+_IDENTIFIER_RE = re.compile(
+    r"-(?![0-9])$"
+    r"|-[a-zA-Z_+*/=><!&][a-zA-Z0-9_\-+*/=><!&]*$"
```
CoolCatMona: `grammar.lark` permits `?` in identifiers, but this regex does not. A name such as `eligible?` parses and then fails variable creation. Please keep this validator aligned with the grammar.

app/utils/scheduling.py
reid23: added `reconcile_tag_backed_match_slots`, runs at the start of live scheduling pass so resolved tags are written before the status setting stuff

app/utils/parser.py
reid23: made it load script vars while doing the skip condition parsing. added check for circular dependency issues

app/routes/tournaments/matches_admin.py
reid23: oops. added recomputation

app/routes/tournaments/matches_admin.py
reid23: cope just made it throw an error if the variable is used.

app/routes/tournaments/matches_admin.py
reid23: extended :+1:

migrations/versions/0014_tag_expressions_script_variables.py
reid23: added explicit deletion for ScriptVariable in delete_tournament and added cascade and adjusted migration accordingly

app/routes/tournaments/read.py
reid23: lwk it *shouldnt* be like this because people should be able to see the logic. but we don't have a good UI for this yet so its fine to just hide it and show that it's a mystery tag.

app/routes/tournaments/read.py
reid23: deferring this to later. we aren't running tournaments that large and nobody will use this except me anyway probably. at least for a little while. cope harder

reid23: see #263

app/utils/parser.py
reid23: fine,,, did this properly

app/utils/toml_helpers.py
reid23: validated that all variables, tags, fields, and matches must be tables in the toml

app/utils/helpers.py
reid23: oopsie, updated

app/utils/parser.py
reid23: updated to allow `?`

